PR #1, opened 2020-06-16
Add enum-like collections

jreut:
FrozenSet and FrozenMap wrap Arrays and Objects in Proxy objects that throw errors on mutation and missing keys.

As an example, I've extracted some strings into a Nation constant.

### DIFF
--- a/frozen.js
+++ b/frozen.js
@@ -1,0 +1,35 @@
+const cleanroom = () => Object.freeze(Object.create(null))
+const id = x => x
+
+const handlerFor = (store, { getter }) => ({
+  get: (_target, property, _receiver) => {
+    if (property === Symbol.iterator) {
+      return store[Symbol.iterator].bind(store)
+    }
+
+    if (store.has(property)) {
+      return getter(property)
+    }
+
+    throw new Error(`"${String(property)}" not found in this enum`)
+  },
+  set: () => { throw new Error("enums cannot be mutated") },
+})
+
+const FrozenSet = (ary) => {
+  const store = new Set(ary)
+
+  return new Proxy(cleanroom(), handlerFor(store, { getter: id }))
+}
+
+const FrozenMap = (obj) => {
+  const store = new Map()
+
+  for (const key of Object.getOwnPropertyNames(obj)) {
+    store.set(key, obj[key])
+  }
+
+  return new Proxy(cleanroom(), handlerFor(store, { getter: store.get.bind(store) }))
+}
+
+module.exports = { FrozenSet, FrozenMap }
--- a/imperial.js
+++ b/imperial.js
@@ -1,3 +1,5 @@
+const { Nation } = require("./constants")
+
 class Imperial {
   static fromLog(log) {
     return new Imperial(log);
@@ -23,7 +25,7 @@ class Imperial {
       return this.rondelActions(this.getNation(this.log));
     } else if (this.lastMoveWasRondelManeuver(lastMove)) {
       switch (lastMove.payload.nation) {
-        case "FR":
+        case Nation.FR:
           const FRLandDestinations = [
             "brest",
             "dijon",
@@ -58,7 +60,7 @@ class Imperial {
             },
             ...parisActions,
           ];
-        case "GB":
+        case Nation.GB:
           return [
             {
               type: "manuever",
@@ -128,7 +130,7 @@ class Imperial {
               payload: { origin: "berlin", destination: "norway" },
             },
           ];
-        case "AH":
+        case Nation.AH:
           const AHLandDestinations = [
             "warsaw",
             "kiev",
@@ -174,7 +176,7 @@ class Imperial {
             ...budapestActions,
             ...viennaActions,
           ];
-        case "IT":
+        case Nation.IT:
           const ITLandDestinations = [
             "naples",
             "tunis",
@@ -218,16 +220,16 @@ class Imperial {
             type: "coexist",
             payload: {
               province: "western mediterranean sea",
-              incumbent: "IT",
-              challenger: "FR",
+              incumbent: Nation.IT,
+              challenger: Nation.FR,
             },
           },
           {
             type: "fight",
             payload: {
               province: "western mediterranean sea",
-              incumbent: "IT",
-              challenger: "FR",
+              incumbent: Nation.IT,
+              challenger: Nation.FR,
             },
           },
           ,
@@ -244,12 +246,12 @@ class Imperial {
 
   nations() {
     let nations = {};
-    ["AH", "IT", "FR", "GB", "GE", "RU"].map((nation) => {
+    for (const nation of Nation) {
       nations[nation] = {
         controller: this.getController(nation, this.log),
         treasury: this.getTreasury(nation, this.log),
       };
-    });
+    };
     return nations;
   }
 
@@ -407,7 +409,7 @@ class Imperial {
         rondelActions[rondelActions.length - 1].payload.nation;
       return this.nextNation(lastTurnNation);
     } else {
-      return "AH";
+      return Nation.AH;
     }
   }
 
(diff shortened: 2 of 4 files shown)


frozen.js
```diff
@@ -0,0 +1,35 @@
...
+
+    throw new Error(`"${String(property)}" not found in this enum`)
+  },
+  set: () => { throw new Error("enums cannot be mutated") },
+})
+
+const FrozenSet = (ary) => {
+  const store = new Set(ary)
+
+  return new Proxy(cleanroom(), handlerFor(store, { getter: id }))
```
Thrillberg, line 22: Am I correct that in our use of `Proxy`, we're treating `cleanroom()` like a throwaway? Do we even need anything here?

frozen.js
```diff
@@ -0,0 +1,35 @@
+const cleanroom = () => Object.freeze(Object.create(null))
+const id = x => x
+
+const handlerFor = (store, { getter }) => ({
+  get: (_target, property, _receiver) => {
+    if (property === Symbol.iterator) {
```
Thrillberg, line 6: I read the docs on [`Symbol.iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator) and it doesn't make sense to me. 😞 Can you explain what `property` is here and when it might equal `Symbol.iterator`?

imperial.js
```diff
@@ -244,12 +246,12 @@ class Imperial {
 
   nations() {
     let nations = {};
-    ["AH", "IT", "FR", "GB", "GE", "RU"].map((nation) => {
+    for (const nation of Nation) {
```
Thrillberg, line 249: Not that this is a good idea, but is it conceivably possible for us to give `FrozenSet` a function like `map` so that we can call `Nation.map((nation) => ...)`? Or maybe make another constant, which is `Nations` (plural) and _that_ refers to the iterable array of them?

imperial.js
```diff
@@ -448,11 +450,18 @@ class Imperial {
   }
 
   nextNation(lastTurnNation) {
-    const nations = ["AH", "IT", "FR", "GB", "GE", "RU"];
-    if (lastTurnNation === "RU") {
-      return "AH";
+    const turnOrder = [
```
Thrillberg, line 453: Thanks for renaming this!

Thrillberg (approved):
Thanks Jordan! I really like this! It all looks great to me, but I left a few questions mostly about things I didn't understand.

Thrillberg:
Also, fyi I've changed the base branch for this repo to be `main` (and added a GitHub action to run tests on all commits!).

Thrillberg:
Whoops! And then I closed `master` and that plorpped your PR. Sorry! Can you re-aim it at `main`?

frozen.js
jreut: This is a great question, and I'll try to answer it as best I can with some analogy to Ruby's object model. Also, I use the nomenclature "ECMAScript" and "ES" here because I want to be an iconoclast.

In the same way `Enumerable` Ruby objects all implement the well-known method `#each`, ECMAScript objects that want to be [iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol) implement a well-known method called (waves hands) `Symbol.iterator`.

Ruby and ES share a "prototypal" inheritance model. Indeed, Ruby's "classical" inheritance is built from this more flexible model. For instance, while Ruby's `Object` does not implement `#each`

```ruby
>> Object.methods.include? :each
=> false
```

We can write a custom implementation for _a single instance_:

```ruby
>> custom = Object.new
?> def custom.each
?>   yield "hello from custom!"
>> end
>> custom.each { |x| "got: #{x}" }
=> "got: hello from custom!"
```

And, perhaps bizzarely, we can make that _single instance_ `Enumerable`:

```ruby
>> custom.singleton_class.include Enumerable
=> #<Class:#<Object:0x00001b3be9574910>>
>> custom.map { |x| "got #{x}" }
=> ["got hello from custom!"]
```

While all other instances of `Object` are unaffected.

```ruby
>> Object.new.map { |x| "this doesn't work" }
Traceback (most recent call last):
        4: from /usr/local/bin/irb27:23:in `<main>'
        3: from /usr/local/bin/irb27:23:in `load'
        2: from /usr/local/lib/ruby/gems/2.7/gems/irb-1.2.3/exe/irb:11:in `<top (required)>'
        1: from (irb):13
NoMethodError (undefined method `map' for #<Object:0x00001b3b6211b5d0>)
Did you mean?  tap
```

---

As you may suspect, we can do the same in ECMAScript.

```es6
// By default, instances of Object are not iterable
> for (x of {}) {
... console.log("got:", x)
... }
Uncaught TypeError: {} is not iterable

// But we can implement a Symbol.iterator generator function.
> custom = {}
{}
> custom[Symbol.iterator] = function* () { yield "hello from custom!" }
[GeneratorFunction]
> for (x of custom) {
... console.log("got:", x)
... }
got: hello from custom!
```

Unfortunately, there's no equivalent of the `Enumerable` module, so we don't get `map` for free. We would need to define our own, perhaps like this:

```es6
> custom.map = function (mapper) { return [...this].map(mapper) }
[Function]
> custom.map(x => `got: ${x}`)
[ 'got: hello from custom!' ]
```

(I admit a little..._elision_ here. The `...` spread operator uses `Symbol.iterator`.)

---

Now for the part about `property` and the `Proxy` API. In Ruby, each object has two maps: one for instance variables and one for methods. This separation facilitates the separate behavior of lookups into each map. Instance variable lookups stop at the object itself, and method lookups walk the inheritance list. ECMAScript objects have only one map: the map of _properties_, and property lookup is the same whether the programmer intends to look for "methods" or "variables." An example:

```es6
> customPrototype = { theAnswer: 42 }
{ theAnswer: 42 }
> childA = Object.create(customPrototype)
{}
> childB = Object.create(customPrototype)
{}
> childA.theAnswer = 108
108
> childA.theAnswer
108
> childB.theAnswer
42
> Object.getPrototypeOf(childA).theAnswer
42
```

---

So, when might `prototype === Symbol.iterator`? It is when we use [`for...of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). The `get()` function of a `Proxy`'s handler object is invoked upon every property access. To tie it all together:

```es6
> customIterator = function* () { yield "helu" }
[GeneratorFunction]
> underlyingObject = { theAnswer: 42 }
{ theAnswer: 42 }
> proxiedObject = new Proxy(underlyingObject, {
... get: function (target, property, receiver) {
..... if (property === Symbol.iterator) {
....... return customIterator
....... }
..... return Reflect.get(...arguments)
..... }
... }
... )
Proxy [ {}, { get: [Function: get] } ]
> underlyingObject.theAnswer = 42
42
> proxiedObject.theAnswer
42
> for (x of underlyingObject) { console.log("got:", x) }
Uncaught TypeError: underlyingObject is not iterable
> for (x of proxiedObject) { console.log("got:", x) }
got: helu
undefined
```

Does that make sense?

frozen.js
jreut: I wrote a giant explanation and then accidentally closed the tab, so I have a pithy one for you.

We want to use an immutable object that has no properties or prototype chain of its own. `null` and `undefined` satisfy the lack of properties and prototype chain, but alas, they are not objects:

```es6
> new Proxy(null, {})
Uncaught TypeError: Cannot create proxy with a non-object as target or handler
> new Proxy(undefined, {})
Uncaught TypeError: Cannot create proxy with a non-object as target or handler
```

Instead, we must resort to creating an empty object with no prototype (`Object.create(null)`) and freeze it to prevent new property addition.

imperial.js
jreut: That actually sounds like a great idea :) I can definitely make a map function. It'll look something like this:

```es6
function map (fn) { [...this].map(fn) }
```

While we're at it, how about a "case analysis" function?

```es6
Nation.switch({
  GB: () => /* ... */,
  FR: () => /* ... */,
  /* remaining cases */
})
```

This function will also perform an exhaustiveness check! That is, it will throw an error if we fail to handle every member of the enum.

frozen.js
jreut: This serves a similar function to Ruby's [`BasicObject`](https://ruby-doc.com/core/BasicObject.html)

PR #2, opened 2020-06-20
Add Enum

jreut:
The entrypoint to this feature is Enum.fromArray([String]).

This feature provides a simple "union" data type. It gives us

- Strict property access

  Attempts to access a nonexistent property (ie. via misspelling) will
  cause an exception instead of returning the undefined value.

- Set membership

  Because each Enum and its instances are immutable, Set membership
  works in the expected way.

- Exhaustive case analysis

  Every instance of an Enum has a #when() function that accepts an
  Object mapping each possible value to a function. This lets us
  write "switch" statements that validate we've handled every possible
  value.

- Iterability

  This lets us, say, fill in a Map whose keys are the members of the
  Enum and whose values are some transformation of them:

      const out = new Map()
      for (const member of Enum.fromArray(["ONE", "THREE"])) {
        out.set(member, member.unwrap().length)
      }

- Stringification

  Debug output will say something like "[object MyEnum.MEMBER]" instead
  of "[object Object]"

### DIFF
--- a/imperial.js
+++ b/imperial.js
@@ -1,3 +1,5 @@
+const { Nation } = require("./constants")
+
 class Imperial {
   static fromLog(log) {
     return new Imperial(log);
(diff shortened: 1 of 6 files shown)


jreut:
This won't be ready until I integrate it with the rest of the project, but I wanted to post what I have so far.

imperial.js
```diff
@@ -522,12 +524,14 @@ class Imperial {
...
-      return nations[nations.indexOf(lastTurnNation) + 1];
-    }
+    return lastTurnNation.when({
+      AH: () => Nation.IT,
+      IT: () => Nation.FR,
+      FR: () => Nation.GB,
+      GB: () => Nation.GE,
+      GE: () => Nation.RU,
+      RU: () => Nation.AH,
+    })
```
jreut, line 534: This may require more lines of code, but check out how much clearer it is!

imperial.js
```diff
@@ -296,15 +298,15 @@ class Imperial {
...
+    for (const nation of Nation) {
+      nations.set(nation, {
         controller: this.getController(nation, this.log),
         treasury: this.getTreasury(nation, this.log),
         taxChartPosition: this.getTaxChartPosition(nation),
         powerPoints: this.getPowerPoints(nation),
-      };
-    });
+      });
+    };
```
jreut, line 309: This is an interim change. I hope to eventually make a "Record" type that is similar to `Enum` in its strictness, except it stores key-value pairs of string-anything instead of storing a set of strings.

PR #5, opened 2020-06-21
Remove Enum#unwrap and add more stringification

jreut:


enum.js
```diff
@@ -1,7 +1,16 @@
 const cleanroom = Object.freeze(Object.create(null))
 
-const prototype = {
-  unwrap () { return this.value },
+class Enum {
```
Thrillberg, line 3: Ah so we _can_ use `class` syntax. Cool!

enum.js
```diff
@@ -1,7 +1,16 @@
...
+  constructor (value, members, label) {
+    Object.defineProperties(this, {
+      value: { value: value, enumerable: true },
+      members: { value: members, enumerable: false },
+      label: { value: label, enumerable: true },
+    })
```
Thrillberg, line 9: So all we're doing here is setting each incoming argument to be an object of itself plus whether it is enumerable? Do we use the enumerable boolean anywhere?

enum.js
```diff
@@ -13,50 +22,43 @@ const prototype = {
...
     }
-  },
-  get [Symbol.toStringTag] () {
-    return `${this.label}.${this.unwrap()}`
-  },
-}
-
-const Enum = function (value, members, label) {
-  me = Object.create(prototype)
+    throw new Error("unreachable")
```
Thrillberg, line 25: lol feels like the failure case for something called `when` should be `"never"`. I don't actually think we should change this, I just think words are funny.

enum.js
```diff
@@ -13,50 +22,43 @@ const prototype = {
...
-    members: { value: members, enumerable: false },
-    label: { value: label, enumerable: false },
-  })
+  get [Symbol.toStringTag] () {
+    return `${this.label}.${this.value}`
+  }
 
-  return Object.freeze(me)
-}
+  static fromArray (ary, label) {
```
Thrillberg, line 32: `static` means class method, right?

Thrillberg (approved):
Looks great to me! No need to answer any of my questions inside, I was just writing comments as a way to process/understand what's going on. But please feel free to correct anything I might have misread. Merge when the spirit strikes! 😄

enum.js
jreut: It's a little different than that. What I wrote here is exactly equivalent to this:

```es6
constructor (value, members, label) {
  this.value = value
  this.members = members
  Object.defineProperty(this, 'members', { enumerable: false })
  this.label = label

  Object.freeze(this)
}
```

I just want the "members" property to not be enumerable so debug tools and pretty printers don't display it.

See: [Object.defineProperties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties)

PR #6, opened 2020-06-21
Add memoize() and use it with action creation functions

jreut:
I've converted actions of type "maneuver" and "bondPurchase" so far within the implementation file.

### DIFF
--- a/imperial.js
+++ b/imperial.js
@@ -1,4 +1,5 @@
-const { Nation } = require("./constants")
+const { Nation } = require("./constants");
+const { bondPurchase, maneuver } = require("./actions");
 
 class Imperial {
   static fromLog(log) {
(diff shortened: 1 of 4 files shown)


memo.js
```diff
@@ -0,0 +1,30 @@
+NO_VALUE = Symbol("NO_VALUE");
+
+class Trie {
```
Thrillberg, line 3: `Trie`?

memo.js
```diff
@@ -0,0 +1,30 @@
...
+      this.children.set(head, new Trie(NO_VALUE));
+    }
+    return this.children.get(head).getOrInsert(tail, fn, args);
+  }
+}
+
+const memoize = (fn) => {
+  const memory = new Trie(NO_VALUE);
+
+  return (...args) => memory.getOrInsert(args, fn, args);
```
Thrillberg, line 27: I'm confused by `...args`. Where are those coming from?

imperial.js
```diff
@@ -29,50 +30,17 @@ class Imperial {
...
-        },
-        {
-          type: "bondPurchase",
-          payload: { nation: Nation.GE, player: "Claudia", cost: 2 },
-        },
-        {
-          type: "bondPurchase",
-          payload: { nation: Nation.RU, player: "Claudia", cost: 4 },
-        },
+        bondPurchase(Nation.AH, "Claudia", 4),
```
Thrillberg, line 33: I love the shorter syntax! I wonder if we might be losing something by not having the arguments have keywords? It might get confusing as we add more actions. What do you think?

Thrillberg (approved):
Looks great to me! I think this can be merged, assuming the keyword-less syntax is what we want to go with. I can be convinced either way!

memo.js
jreut: [Trie!](https://xlinux.nist.gov/dads/HTML/trie.html)

memo.js
jreut: Much like this Ruby syntax:

```ruby
def blorp(*args)
  args.each_with_index { |e, i| puts "##{i}: #{e}" }
end
blorp("hello", "nice", "world")
#=> #0: hello
#=> #1: nice
#=> #2: world
```

The `...args` syntax in ES collects all of the function's arguments into an array.

```es6
const blorp = (...args) => { args.forEach((e, i) => { console.log(`#${i}: ${e}`) }) }
blorp("hello", "nice", "world")
//=> #0: hello
//=> #1: nice
//=> #2: world
```

imperial.js
jreut: I'm definitely up to use keyword-style arguments! I'll make that change.

memo.js
Thrillberg: I like how 50% of that article was about how to pronounce it.

memo.js
Thrillberg: Hmm ok but in the context of the `memoize` constant, where do they come from? The `fn` on line 24?

Thrillberg:
I just realized that `maneuver` actions need to have another piece of data: the nation whose pieces are moving. This is because two nations can share a province with their armies or fleets and if one of them evacuates, we'll need to know which one moved. What do you think of making `maneuver` look like this instead?

```js
{
    type: "maneuver",
    payload: { nation: Nation.FR, origin: "portugal", destination: "spain" },
}
```

On second thought, I suppose it's not _strictly_ necessary to include nation since we can infer which nation is moving based on who most recently went to the maneuver rondel slot. But that is trickier than just encoding the nation into the action itself. What do you think?

Thrillberg:
Actually I take it back. I think we can leave the log entry for maneuver as-is and just write logic to figure out which maneuvers belong to whomb.

jreut:
This is superceded by #8

PR #9, opened 2020-06-24
Refactor to tick

Thrillberg:


imperial.js
```diff
@@ -3,23 +3,50 @@ const Action = require("./action");
...
     };
   }
 
+  setupNations() {
+    let nations = {};
+    [Nation.AH, Nation.IT, Nation.FR, Nation.GB, Nation.GE, Nation.RU].forEach(
+      (nation) => (nations[nation] = { controller: "", treasury: 0 })
+    );
+    return nations;
+  }
```
jreut: What is the output of this function? In the words of Inigo Montoya, "I do not think it means what you think it means." By way of abstract illustration, here's what I get when I do something similar with another Enum:

```es6
const { Enum } = require("./enum")
const Test = Enum.fromArray(["SPAM", "EGGS"])

const obj = [...Test].reduce((acc, e) => ({...acc, [e]: `got ${e.value}`}), {})
//=> {
//=>   '[object <anonymous>.SPAM]': 'got SPAM',
//=>   '[object <anonymous>.EGGS]': 'got EGGS'
//=> }
obj[Test.SPAM]
//=> "got SPAM"
obj[String(Test.SPAM)]
//=> "got SPAM"
Test.SPAM.value
//=> "SPAM"
obj[Test.SPAM.value]
//=> undefined
```

This code seems to work, although accidentally. As you may know, keys of objects are silently stringified in ECMAScript. `Enum` instances [implement stringification](https://github.com/Thrillberg/imperial/blob/54bd37439b77e7a73ca3392b6eb820f84856a7de/enum.js#L28-L30) as `${label}.${tag}`, which is what we see in the keys of our `obj`. (That extra `[object ...]` stuff comes from the ES interpreter).

I ran into this problem myself while first integrating the `Nation` enum, and I [solved it with a Map](https://github.com/Thrillberg/imperial/blob/54bd37439b77e7a73ca3392b6eb820f84856a7de/imperial.js#L352-L363). What do you think about converting this method to use a Map instead of an Object?

imperial.js
```diff
@@ -3,23 +3,50 @@ const Action = require("./action");
...
+        (player) => (this.players[player] = { cash: 13 })
+      );
+    } else if (action.type === "assignStartingNation") {
+      this.nations[action.payload.nation].controller = action.payload.player;
+    } else if (action.type === "bondPurchase") {
+      this.nations[action.payload.nation].treasury += action.payload.cost;
+      this.players[action.payload.player].cash -= action.payload.cost;
+    }
+    this.log.push(action);
+  }
```
jreut: If we change `this.nations` to be a Map, we'll need to change our usage to `this.nations.get(action.payload.nation)`.

schnelleinsteig.test.js
```diff
@@ -130,119 +130,193 @@ describe("Schnelleinsteig", () => {
...
+        Action.assignStartingNation({ nation: Nation.RU, player: "Daniel" }),
+        Action.assignStartingNation({ nation: Nation.FR, player: "Claudia" }),
+        Action.assignStartingNation({ nation: Nation.GB, player: "Bert" }),
+        Action.assignStartingNation({ nation: Nation.IT, player: "Anton" }),
+        Action.bondPurchase({ nation: Nation.AH, player: "Claudia", cost: 2 }),
+        Action.bondPurchase({ nation: Nation.IT, player: "Anton", cost: 9 }),
+        Action.bondPurchase({ nation: Nation.FR, player: "Claudia", cost: 9 }),
         Action.bondPurchase({ nation: Nation.FR, player: "Daniel", cost: 2 }),
+        Action.bondPurchase({ nation: Nation.GB, player: "Anton", cost: 2 }),
+        Action.bondPurchase({ nation: Nation.GB, player: "Bert", cost: 9 }),
```
jreut: How can we avoid this (and the previous) copy-pasting?

imperial.js
Thrillberg: Ah ok that almost makes sense. I think I'll screw this up a couple more times before I get it!

Thrillberg:
@jreut Thanks for the review! Ready for another round, methinks!

jreut (approved):
lgtm!

PR #12, opened 2020-07-05
Add index.html and Vue

Thrillberg:


Thrillberg:
@jreut I'm happy to merge this without your review if you'd prefer but just wanted to let you know that I think this is a reasonable place to check in some work. I decided against Choo in favor of Vue since I think there will be more docs/help and I was having trouble getting Choo installed using ES6 modules. Vue just seemed easier to get off the ground. What do you think?

constants.js
```diff
@@ -1,5 +1,5 @@
-const { Enum } = require("./enum");
+import { Enum } from "./enum.js";
```
jreut, line 1: Pardon my ignorance on this, but do we prefer to include the `.js` extension in `import` statements?

jest.config.js
```diff
@@ -129,7 +129,7 @@ module.exports = {
   // snapshotSerializers: [],
 
   // The test environment that will be used for testing
-  testEnvironment: "node",
+  testEnvironment: "jest-environment-node",
```
jreut: I'm curious: is this change strictly necessary?

I tried reverting this change locally and tests did pass.

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "lint": "prettier -c *.js",
     "lint:fix": "prettier --write *.js",
-    "test": "jest"
-  }
+    "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js"
+  },
+  "type": "module"
 }
```
jreut, line 14: Could we add a development dependency on `static-server` and include a ["start"](https://docs.npmjs.com/cli/start.html) script that runs it?

schnelleinsteigLog.js
```diff
@@ -0,0 +1,104 @@
...
+  }),
+  Action.maneuver({
+    origin: "moscow",
+    destination: "turkey",
+  }),
+  Action.maneuver({
+    origin: "moscow",
+    destination: "lemberg",
+  }),
+];
```
jreut, line 104: 🙏🏼

main.js
```diff
@@ -0,0 +1,53 @@
...
+      </div>
+    </div>
+  </li>
+  `,
+});
+
+Vue.component("bond", {
+  props: ["nation", "cost"],
+  template: `<li class="bond">{{ nation }}{{ cost }}</li>`,
+});
```
jreut, line 30: I'm unfamiliar with Vue. How would we have written this in React+JSX? Like this?

```es6
const bond = ({ nation, cost }) => <li class="bond">{nation}{cost}</li>
```

jreut (approved):
lgtm!

constants.js
Thrillberg: Nope, you're right! I will remove it.

jest.config.js
Thrillberg: Yeah this is now working for me too. Great catch! Thanks!

constants.js
Thrillberg: Actually when I removed those, it broke on the front end I think because it was looking for `"text/plain"` files, not JS files.

main.js
Thrillberg: I'm not sure. It's been so long since I've done React and I never did do it in the newer function-based syntax.

PR #15, opened 2020-07-11
Use an initial state

jreut:


constants.js
```diff
@@ -1,5 +1,6 @@
 import { Enum } from "./enum.js";
 
-const Nation = Enum.fromArray(["AH", "IT", "FR", "GB", "GE", "RU"], "Nation");
-
-export { Nation };
+export const Nation = Enum.fromArray(
+  ["AH", "IT", "FR", "GB", "GE", "RU"],
+  "Nation"
+);
```
Thrillberg: Thanks for tightening this up!

imperial.js
```diff
@@ -113,30 +113,39 @@ export default class Imperial {
...
-      const maneuverActions = this.log.filter(
-        (action) =>
-          action.type === "rondel" &&
-          (action.payload.slot === "maneuver1" ||
-            action.payload.slot === "maneuver2")
-      );
-      const lastManeuver = maneuverActions[maneuverActions.length - 1];
-      this.provinces.get(action.payload.destination).flag =
-        lastManeuver.payload.nation;
+    switch (action.type) {
```
Thrillberg: Do you think we may ultimately want something other than a `switch` here? I'm not sure what would be better but I guess if we want to do some OO stuff we could embellish `action`s with their own `doTheThing` methods? I don't necessarily have any opposition to the `switch` statement, I just want to be more aware of the tradeoffs.

constants.js
```diff
@@ -1,5 +1,84 @@
...
+  [Nation.GE, Nation.IT],
+  [Nation.RU, Nation.FR],
+  [Nation.AH, Nation.GE],
+  [Nation.IT, Nation.GB],
+  [Nation.FR, Nation.AH],
+  [Nation.GB, Nation.RU],
+].reduce((map, [k, v]) => {
+  map.set(k, v);
+  return map;
+}, new Map());
```
Thrillberg: I'm just curious, might it be preferable/more readable to initialize the new `Map` directly with the arguments rather than doing a `reduce`?
```javascript
export const NationCard = new Map(
  [
    [Nation.GE, Nation.IT],
    [Nation.RU, Nation.FR],
    [Nation.AH, Nation.GE],
    [Nation.IT, Nation.GB],
    [Nation.FR, Nation.AH],
    [Nation.GB, Nation.RU],
  ]
);
```

constants.js
```diff
@@ -1,5 +1,84 @@
...
+
+/* Placeholder for string-valued "enums" before we migrate them to real
+ * Enum instances
+ */
+export const Strings = {
```
Thrillberg: Great move!

imperial.js
```diff
@@ -1,20 +1,62 @@
-import { Nation } from "./constants.js";
+import { Nation, Strings, NationCard } from "./constants.js";
```
Thrillberg: Suuuuuper small but how do you feel about alphabetizing these?
```javascript
{ Nation, NationCard, Strings }
```

imperial.js
```diff
@@ -1,20 +1,62 @@
...
 
+Map.fromIterable = (iter, getValue) => {
+  const m = new Map();
+  for (const k of iter) {
+    m.set(k, getValue(k));
+  }
+  return m;
+};
+
+Array.fromIterable = (iter, getValue) => {
```
Thrillberg: I _think_ I like these `fromIterable` helpers but I wonder if they should go in a different file? Maybe a `helpers` file since that seems to me like what they are? Or maybe it's premature to move them out now.

imperial.js
```diff
@@ -1,20 +1,62 @@
...
+  for (const k of iter) {
+    a.push(getValue(k));
+  }
+  return a;
+};
+
 export default class Imperial {
-  static fromLog(log) {
-    let game = new Imperial();
+  static fromLog(log, initialState) {
```
Thrillberg: How do you see `initialState` being most useful as a thing to pass in here? What does this enable for us?

imperial.js
```diff
@@ -1,20 +1,62 @@
...
+    }));
+
+    if (initialState !== undefined) {
+      this.order = initialState.players.map((p) => p.id);
+      for (const { id, nation } of initialState.players) {
+        this.players[id] = {
+          name: id,
+          bonds: [
+            { nation, cost: 9 },
+            { nation: NationCard.get(nation), cost: 2 },
```
Thrillberg: I'm not sure how I feel about `get`ting a nation off a `NationCard` as being the 2-cost bond one but I can't think of a better alternative so I guess it's fine. Alternatives I can think of suffer from being more verbose/holding more data.

I do wonder how you feel about making constants out of `Bond`s. Do you think that makes sense? I think memoizing them could make sense.

imperial.test.js
```diff
@@ -0,0 +1,113 @@
...
+        players: [
+          { id: "a", nation: Nation.RU },
+          { id: "b", nation: Nation.FR },
+          { id: "c", nation: Nation.GB },
+          { id: "d", nation: Nation.AH },
+          { id: "e", nation: Nation.IT },
+          { id: "f", nation: Nation.GE },
+        ],
+      };
+      const sut = new Imperial(state);
```
Thrillberg: What is `sut`?

imperial.test.js
```diff
@@ -0,0 +1,113 @@
+import Imperial from "./imperial";
+import { Nation } from "./constants.js";
+
+describe("Imperial", () => {
+  describe("constructor", () => {
+    describe("six players", () => {
```
Thrillberg: Could this be straightforwardly done for 5, 4, 3, and 2 player games too? Is it worth writing out tests for those too?

Thrillberg (approved):
This is AWESOME!! Thank you so much for writing this code. I'm very excited to get it merged! I have some questions and observations that I'd love to hear more about.

constants.js
jreut: Oop, I didn't know that was possible! This changes everything

jreut:
Thanks for leaving all those comments :) Although I won't reply to all of them I do agree with most of them; this code is very rough as is. In my next iteration I'll keep the original goal of test-driving this idea of constructing an `Imperial` instance with an initial state. An initial state seems to me a better abstraction than the idea of "setup" actions and "gameplay" actions. What do you think?

imperial.test.js
jreut: System Under Test

but i think that's probably more confusing than `game`

Thrillberg:
>An initial state seems to me a better abstraction than the idea of "setup" actions and "gameplay" actions.

Mmm yes that makes a lot of sense. I think I didn't quite understand that goal in mind from the start so yeah that sounds great!

Also, I think it might be better to aim to merge this into `main` when it's ready, rather than `more-exploratory-vue-stuff`. I think I may use that branch to spike on and pull off merge-ready items as they form à la that other PR I just merged. What do you think?

imperial.test.js
jreut: Yes, I am doing the 5, 4, 3 and 2 cases presently!

jreut:
This one is a doozy. If you like, I am up to talk it through over the phone :)

schnelleinsteig.test.js
```diff
@@ -15,145 +15,41 @@ const rondelSlots = [
 ];
 
 describe("Schnelleinsteig", () => {
-  describe("setup for four players", () => {
-    test("All players receive 13 million", () => {
-      const log = [
-        Action.playerSeating({ order: ["Daniel", "Claudia", "Bert", "Anton"] }),
-      ];
+  describe.only("setup for four players", () => {
```
Thrillberg: No `only` pl0x.

imperial.test.js
```diff
@@ -0,0 +1,375 @@
...
+        });
+      });
+
+      test("nations", () => {
+        const expected = new Map(
+          [
+            [Nation.RU, "a", 11],
+            [Nation.FR, "b", 11],
+            [Nation.GB, "c", 11],
+            [Nation.GE, "d", 9], // AH controller would have bought GE 2
```
Thrillberg: Improve comment plz

schnelleinsteigLog.js
```diff
@@ -2,20 +2,29 @@ import { Nation } from "./constants.js";
...
+    type: "init",
+    payload: {
+      players: [
+        { id: "Daniel", nation: Nation.RU },
+        { id: "Claudia", nation: Nation.FR },
+        { id: "Bert", nation: Nation.GB },
+        { id: "Anton", nation: Nation.IT },
+      ],
+    },
+  },
```
Thrillberg: Can this be an `Action`?

imperial.js
```diff
@@ -112,13 +106,16 @@ export default class Imperial {
...
 
   tick(action) {
-    if (action.type === "playerSeating") {
-      this.seatPlayers(action);
-    } else if (action.type === "assignStartingNation") {
-      this.assignStartingNation(action);
+    this.log.push(action);
+
+    if (action.type === "init") {
+      Object.assign(this, setup(action.payload));
```
Thrillberg: Explicitly use `this.players = ...` and `this.nations = ...` etc.




pl00z

schnelleinsteig.test.js
jreut: Removing this makes all the Schnelleinsteig tests fail 😿 

I think I'll remove all the changes I made to Schnelleinsteig, especially since I don't want to step on your toes.

imperial.js
```diff
@@ -112,7 +114,13 @@ export default class Imperial {
   }
 
   tick(action) {
-    if (action.type === "playerSeating") {
+    if (action.type === "init") {
+      const { players, order, nations } = setup(action.payload);
+      this.players = players;
+      this.order = order;
+      this.nations = nations;
+      return;
```
jreut, line 122: We don't want to run any side effects that may happen in the rest of `#tick`, so we exit early.

jreut (commented):
I am ready for another round of review! I cut out the parts that change the Schnelleinsteig for now.

action.js
```diff
@@ -35,6 +35,7 @@ const makeAction = (type, payloadKeys) => {
 };
 
 export default {
+  init: makeAction("init", ["players"]),
```
Thrillberg: This is about as nitpicky as it gets but how do you feel about changing this to `initialize`? It's purely aesthetic for me but I prefer to see full words, I guess. 😄

enum.js
```diff
@@ -19,7 +19,7 @@ class Enum {
     }
     for (const c of Object.getOwnPropertyNames(cases)) {
       if (Object.is(this.value, c)) {
-        return cases[c]();
+        return cases[c](this);
```
Thrillberg, line 22: I think I meant to ask about this before but what necessitated this change?

imperial.js
```diff
@@ -12,6 +13,7 @@ export default class Imperial {
     this.log = [];
     this.nations = this.setupNations();
     this.players = {};
+    this.order = null;
```
Thrillberg, line 16: This is just idle curiosity on my part but is it preferable to use `null` over `[]` in this case? Should we initialize `this.log` to be `null` too? And `this.players`? Or is it just taste?

imperial.js
```diff
@@ -112,7 +114,13 @@ export default class Imperial {
...
 
   tick(action) {
-    if (action.type === "playerSeating") {
+    if (action.type === "init") {
+      const { players, order, nations } = setup(action.payload);
+      this.players = players;
+      this.order = order;
+      this.nations = nations;
+      return;
+    } else if (action.type === "playerSeating") {
```
Thrillberg, line 123: I wonder if `playerSeating` should occur in `setup` too? Or save it for later to think about another day?

imperial.test.js
```diff
@@ -0,0 +1,376 @@
+import Imperial from "./imperial";
+import { Nation } from "./constants.js";
+import Action from "./action";
+
+describe("Imperial", () => {
+  describe("constructor", () => {
+    describe("six players", () => {
+      const state = {
```
Thrillberg: How do you feel about naming this variable `payload` instead? I think a name like that would help me keep in mind what type of thing this is. `state` doesn't sit right with me because it's really just one small chunk of what may be considered the game's state.

Alternatively, I guess we could name it `players` and make it refer to the array (not the object) and then we'd call it on line 18 below thusly:
```javascript
const game = Imperial.fromLog([Action.init({ players } )]);
```

The more I think about it, the more I like my second suggestion above. What do you think?

imperial.test.js
```diff
@@ -0,0 +1,376 @@
...
+            k,
+            { controller: v, treasury: 11, rondelPosition: null },
+          ])
+        );
+        expect(game.nations).toEqual(expected);
+      });
+    });
+
+    describe("five players", () => {
+      const state = {
```
Thrillberg: Same question about `state` here as above.

imperial.test.js
```diff
@@ -0,0 +1,376 @@
...
+            nation,
+            { controller, treasury, rondelPosition: null },
+          ])
+        );
+        expect(game.nations).toEqual(expected);
+      });
+    });
+
+    describe("four players", () => {
+      const state = {
```
Thrillberg: Same question about `state` here as above.

imperial.test.js
```diff
@@ -0,0 +1,376 @@
...
+            nation,
+            { controller, treasury, rondelPosition: null },
+          ])
+        );
+        expect(game.nations).toEqual(expected);
+      });
+    });
+
+    describe("three players", () => {
+      const state = {
```
Thrillberg: Same question about `state` here as above.

imperial.test.js
```diff
@@ -0,0 +1,376 @@
...
+            nation,
+            { controller, treasury, rondelPosition: null },
+          ])
+        );
+        expect(game.nations).toEqual(expected);
+      });
+    });
+
+    describe("two players", () => {
+      const state = {
```
Thrillberg: Same question about `state` here as above.

setup.js
```diff
@@ -0,0 +1,108 @@
+import { Nation } from "./constants";
+
+const error = (want) => (x) => {
+  throw new Error(`got=${x.value}, want=${want}`);
+};
+
+export default ({ players }) => {
```
Thrillberg, line 7: It's totally fine if this is more effort than it's worth but perhaps our `imperial.test.js` file is really testing `setup.js`? Do you think it would take an inordinate amount of effort to re-purpose `imperial.test.js` to test the setup? And if we do that would we still need to test that `Imperial`'s constructor works?

Ultimately, I'm sure a done thing is a good thing but I'm just curious about the various merits and demerits to each approach.

setup.js
```diff
@@ -0,0 +1,108 @@
...
+        ],
+        FR: () => [
+          { id, nation: Nation.FR },
+          { id, nation: Nation.GE },
+        ],
+        GB: error("AH|IT|FR"),
+        RU: error("AH|IT|FR"),
+        GE: error("AH|IT|FR"),
+      }),
+    2: ({ id, nation }) =>
```
Thrillberg: Should we swap the positions of `2` and `3` so that they run 2-6 in order? Suuuuper small comment 😄

setup.js
```diff
@@ -0,0 +1,108 @@
...
+        n,
+        { treasury: 0, controller: null, rondelPosition: null },
+      ])
+    ),
+    order: players.map((p) => p.id),
+  };
+
+  players
+    .map(nationAssignments[players.length])
+    .flat()
```
Thrillberg, line 58: Wait, `flat()` works?? I thought this was the stuff of legends...

setup.js
```diff
@@ -0,0 +1,108 @@
...
+      .sort(({ cost: aCost }, { cost: bCost }) =>
+        aCost < bCost ? 1 : aCost > bCost ? -1 : 0
+      );
+
+    that.nations.get(n).controller = (byNation[0] || { owner: null }).owner;
+    that.nations.get(n).treasury = byNation.reduce(
+      (sum, { cost }) => sum + cost,
+      0
+    );
+  }
```
Thrillberg: I don't really understand what this is all doing. Can we use some intermediate, meaningfully-named variables?

Thrillberg (commented):
Alright! Great work, Jordan! Thank you for doing all this. I've left many comments but most of them are just navel-gazey fluff. I can definitely turn down my nitpick-o-meter if you want but I just find it so fun to read all this code and be able to think out loud about it!

enum.js
jreut: This change, which is tested [here](https://github.com/Thrillberg/imperial/pull/15/files#diff-3c42b63f302a4effa1f7f78aa92faa94R75-R82), is for now just a convenience to let us use functions like [these](https://github.com/Thrillberg/imperial/pull/15/files#diff-fcd5628e27bdebc81efd345a905292c3R23-R25) in the branches of each `when`. Although it's just for convenience now, later it will become the foundation of a "tagged union" construct, the details of which are pretty fuzzy to me.

imperial.js
jreut: `this.log` is a little different from `this.order`, in that we expect to be able to `this.log.push()`. We don't, however, expect to do anything with `this.order` except to iterate over it. Indeed, if we haven't initialized `this.order`, we really want `this.order.forEach(...)` or `for (const player in this.order) {...}` to fail. Setting it to `null` will give us just that behavior. Make sense?

imperial.js
jreut: We _do_ perform the `Action.playerSeating(...)` steps in `setup()`!

https://github.com/Thrillberg/imperial/blob/144b0c1dbb965f23526e03af56ab79fceff18d5b/imperial.js#L168-L173

We set both `this.order` and `this.players`, so this function is not needed. We do, however, have to wait until we remove all the setup-like actions from Schnelleinsteig.

setup.js
jreut: We're [living in the future](https://caniuse.com/#feat=array-flat), baby

setup.js
jreut: I'll rename `imperial.test.js` to `setup.test.js` and slightly modify the scope of those tests. I think we can test the integration between `setup` and `imperial` when we replace the setup-like actions in the Schnelleinsteig. Sound good?

setup.js
jreut: It is done!

setup.js
jreut: You got it

action.js
jreut: Sure thang

jreut (commented):
All set, and ready for another round!

imperial.js
Thrillberg: Yes, perfect! Thank you!

setup.js
Thrillberg: Yep, sounds great to me!

Thrillberg (approved):
I love it! Thanks for tolerating all my comments. I imagine we might want to re-purpose some of that (very well explained!) bond logic to determine who has controlling interests in a nation since that can change throughout the game. But let's save that for later and just keep in mind that we have the logic here in Setup!

I think we can merge this 🐶!

PR #17, opened 2020-07-12
Indicate the current player

Thrillberg:


jreut (approved):
this is some rock-solid interface design

PR #21, opened 2020-07-17
Add GameBoard

Thrillberg:


gameBoard.js
```diff
@@ -1,64 +1,50 @@
+import Neighbors from "./neighbors";
+
 export default class GameBoard {
   constructor({ nodes, edges }) {
     this.graph = new Map();
     this.byNation = new Map();
 
+    this.setupGraph(nodes);
+    this.setImmediateNeighbors(edges);
```
Thrillberg, line 9: I wonder if this belongs in `neighborsFor()` instead?

jreut (approved):
This looks great. I can understand it, and it is very pleasing that it supports convoying. I'm sure we could continue to change this code but i say :shipit:

PR #22, opened 2020-07-20
Fix imports so that browser does not complain

Thrillberg:
Without the .js suffix I was getting errors about loading the modules
and that their file type ("text/plain", I think?) was wrong.


jreut (approved):
🤷🏼‍♀️

PR #23, opened 2020-07-20
Use setup in schnelleinsteig

Thrillberg:


Thrillberg:
@jreut I've added a `Bond` constant after a lot of back-and-forth with much more complicated constructions than what I ended up with in the PR. What do you think? I feel like it's not as useful/ergonomic as might be desirable but I'm not even sure how we'll _want_ to access our `Bond`s.

PR #26, opened 2020-07-24
Use Bond constants in Setup

Thrillberg:


Thrillberg:
@jreut This piece is ready for review. It is more prep for using `initialize` (and 13 `noop`s) in the `Schnelleinsteig` test.

setup.test.js
```diff
@@ -83,13 +64,28 @@ describe("Imperial", () => {
...
             [Nation.GE, "f"],
-          ].map(([k, v]) => [
-            k,
-            { controller: v, treasury: 11, rondelPosition: null },
-          ])
+          ].map(([k, v]) => {
+            const nationBonds = [...AllBonds].filter(
+              (bond) =>
+                bond.nation === k && bond.number !== 4 && bond.number !== 1
+            );
```
jreut: This code makes me think we should keep all available bonds in one place instead of distributing them among the nations. If we do that, we can instead write a test like this for each case:

```es6
describe('four players', () => { describe('available bonds', () => {
  expected = new Set([...AllBonds]) // clone AllBonds
  for (const n of Nation) {
    n.when({
      IT: () => {
        expected.delete(Bond(n, 1))
        expected.delete(Bond(n, 4))
      },
      FR: () => {
        // Nobody bought FR 1
        expected.delete(Bond(n, 4))
      },
      AH: () => {
        expected.delete(Bond(n, 1))
        expected.delete(Bond(n, 4))
      },
      GE: () => {
        expected.delete(Bond(n, 1))
        expected.delete(Bond(n, 4))
      },
      GB: () => {
        expected.delete(Bond(n, 1))
        // Nobody bought GB 4
      },
      RU: () => {
        // Nobody bought RU 1
        // Nobody bought RU 4
      },
    })
  }
  expect(actual.availableBonds).toEqual(expected)
})})
```

That is clearer in my opinion. We can also use this pool of bonds directly when we generate actions for a player based on bond cost:

```es6
[...game.availableBonds].filter(b => b.cost <= player.cash)
```

What do you think?

jreut (approved):
one comment, but otherwise lgtm!

constants.js
```diff
@@ -19,5 +19,11 @@ export const Bond = memoize((nation, number) => ({
   number,
   cost: cost[number],
 }));
+export const AllBonds = new Set();
+["AH", "IT", "FR", "GB", "GE", "RU"].forEach((nation) =>
+  Object.keys(cost).map((number) =>
+    AllBonds.add(Bond(Nation[nation], parseInt(number)))
+  )
+);
```
jreut: It seems every time we use `AllBonds`, we are cloning it via `[...AllBonds]`. Why not turn this into a function?
```suggestion
export const AllBonds = () => new Set(
["AH", "IT", "FR", "GB", "GE", "RU"].map((nation) =>
  Object.keys(cost).map((number) =>
    Bond(Nation[nation], parseInt(number))
  )
).flat());
```

PR #28, opened 2020-07-26
wip on more Schnelleinsteig

Thrillberg:


setup.js
```diff
@@ -134,6 +134,7 @@ export default ({ players }) => {
       rondelPosition: null,
       flagCount: 0,
       unitCount: 0,
+      units: [],
```
Thrillberg: I want this to be an array (and not a set) because there can be an arbitrary number of "identical" units on one province. Is array a good data type for that?

jreut: For one, what reasons are there to store units on the nation? Why not store them on some representation of provinces directly? Or perhaps store all units in a global array?

In any case, let's say this array has type `Array<Unit>`. What is the type of Unit? something like this pseudocode?

```
type Unit =
  { nation: Nation
  , kind: "army" | "fleet"
  , province: Province
  }
const Array<Unit> units = []
```

What operations do we want on `units`? Perhaps counting how many of a given nation occupy a given province:

```
const strengthOfNationInProvince = (nation: Nation, province: Province) => units
  .filter(u => u.nation === nation && u.province === province)
  .length
```

Pretty straightforward. But how about resolving conflicts? This one is trickier with an array:

```
const fight = (province: Province, aggressor: Nation, defender: Nation) => {
  const aCount = strengthOfNationInProvince(aggressor, province)
  const dCount = strengthOfNationInProvince(defender, province)

  const reduction = Math.min(aCount, dCount)
  for (const i = 0; i < reduction; i++) {
    const aIndex = units.findIndex(u => u.nation === aggressor && u.province === province)
    units.splice(aIndex, 1)
    const dIndex = units.findIndex(u => u.nation === defender && u.province === province)
    units.splice(dIndex, 1)
  }
}
```

It might be easier with another data structure, perhaps a global "bag" of units:

```
type AllUnits = Map<Province, Units>
type Units = Map<Nation, Presence>
type Presence = { fleets: number, armies: number }
const AllUnits units = new Map(/* initialized with all provinces */)
```

With this structure, we get a great simplification:

```
const strengthOfNationInProvince = (nation: Nation, province: Province) => {
  const p = units.get(province).get(nation)
  return p.fleets + p.armies
}
const fight = (province: Province, aggressor: Nation, defender: Nation) => {
  const a = units.get(province).get(aggressor)
  const d = units.get(province).get(defender)
  const reduction = Math.min(a.fleets + a.armies, d.fleets + d.armies)
  for (const i = 0; i < reduction; i++) {
    if (a.armies > 0) { a.armies-- } else { a.fleets-- }
    if (d.armies > 0) { d.armies-- } else { d.fleets-- }
  }
}
```

Thrillberg: Ah this makes sense to me! I've replaced the commits with what I had that was salvageable and otherwise I will start over using something more like what you propose.

Thrillberg: @jreut I've added a `provinces.js` file with a setup of my interpretation of your comment. Does it look right? I also wrote a test but it's pretty much copy-paste of the implementation so I'm not sure if it's useful!

PR #29, opened 2020-07-29
Revise args for GameBoard.neighborsFor

Thrillberg:


Thrillberg:
@jreut ready for review. I achieved enlightenment with this PR.

standardGameBoard.js
```diff
@@ -0,0 +1,73 @@
+import { Nation } from "./constants";
+
+export default {
```
jreut: What if the default export of this module were an instance of `GameBoard` instead of the arguments required to construct it?

jreut (approved):
🕴🏼

PR #32, opened 2020-08-03
Add unit type to import Action

Thrillberg:


Thrillberg:
@jreut This change is necessary because coastal provinces (like Trieste) can import either an army or a fleet, based on the user's choice.

PR #33, opened 2020-08-05
Blimport

Thrillberg:


action.js
```diff
@@ -23,27 +23,33 @@ import { memoize } from "./memo.js";
...
+      });
+      return { type, payload };
     });
-    return { type, payload };
-  });
-  return (obj) => memoized(...payloadKeys.map((k) => obj[k]));
+    return (obj) => memoized(...payloadKeys.map((k) => obj[k]));
+  } else {
+    return (ary) => {
+      return { type, payload: ary };
```
Thrillberg, line 38: @jreut This might be a little hacky. I avoided memoizing here because it was a deep rabbit hole that seemed endless for me. Seems like we might _want_ to memoize each object inside the outermost array but I also don't honestly know why it's vitally important to do. What do you think?

standardSetup.js
```diff
@@ -163,5 +162,26 @@ export default ({ players }) => {
...
+  const provinces = new Map();
+  const armaments = ["vienna", "budapest", "paris", "berlin", "rome", "moscow"];
+  const shipyard = [
+    "bordeaux",
+    "london",
+    "liverpool",
+    "hamburg",
+    "naples",
+    "odessa",
+  ];
```
jreut, line 174: Is this the reason for the split between "standard" setup and "minimal" setup? If that's the case, I think we can do better! We can pass two sets of province names (one for armaments and one for shipyards) to the setup function itself instead of hard-coding it here.

imperial.test.js
```diff
@@ -0,0 +1,104 @@
...
+    Action.initialize({
+      players: [
+        { id: "player1", nation: "nation" },
+        { id: "player2", nation: "nation2" },
+      ],
+      setup,
+    })
+  );
+  return game;
+};
```
jreut, line 37: 👍🏼

imperial.js
```diff
@@ -636,36 +617,16 @@ export default class Imperial {
...
-      out.add(Action.import({ province: "odessa", unit: "fleet" }));
+    const out = new Set();
+    for (const province of this.board.byNation.get(nation)) {
+      if (this.board.graph.get(province).factoryType === "shipyard") {
+        out.add(Action.import({ province, unit: "fleet" }));
+      }
+      out.add(Action.import({ province, unit: "army" }));
     }
     return out;
   }
```
jreut, line 628: 😍

action.js
jreut: I think this is fine for now, but we'll run into trouble when we try to construct sets of actions made in this way. A thought: instead of adding a conditional here, why not write another function, makeNonMemoizedAction (or similar)?

action.js
```diff
@@ -23,27 +23,33 @@ import { memoize } from "./memo.js";
...
+    };
+  }
 };
 
 const noop = Object.freeze({ type: "noop" });
 
 export default {
   noop,
-  initialize: makeAction("initialize", ["players"]),
+  initialize: makeAction("initialize", ["players", "setup"]),
```
jreut, line 47: While it does work, passing a function to the initialize action will likely make our future lives difficult. We want to treat this log as an immutable record of the game's progress. On top of that immutability, we will certainly want to be able to send serialized representations of log entries across the network so other participants can append actions to their logs. So far we have preserved this property of serialization by using "flat" data structures. How might we serialize functions, though? By default they are not represented in JSON:

```es6
> JSON.stringify({huhu: 42, f: x => x + 1})
'{"huhu":42}'
```

I propose we try to keep functions (and other "rich" data structures like class instances) out of these actions. Does that make sense?

jreut (commented):
I have some musings, but considering I'm coming in after the merge and I have no code to illustrate alternative ideas, this all LGTM :)

This is a great step forward in removing hard-coded logic from class Imperial.

PR #37, opened 2020-08-08
Show off a rendered game board

jreut:
I don't intend for this to be merged as-is, but I wanted to show you [what I've been doing](https://github.com/Thrillberg/imperial/blob/555eb6e27023dd6dc2720175ed950342bb6322e0/board.svg) this evening :)


jreut:
TODO: for provinces like naples, imported fleets may only move to westernmediterraneansea. convoys, however, may come from ioniansea.

Thrillberg:
@jreut I'd like to reconsider this commit and see if we want to just merge it as-is, perhaps with some cleanup. I haven't looked at this code in awhile so I don't know what monsters lurk within.

PR #39, opened 2020-08-08
Betterneuver

Thrillberg:


jreut (approved):
lgtm!

Related to these changes, I think my next tedium task is to turn the province strings strewn about into an Enum.

PR #54, opened 2020-08-23
Try vue cli

Thrillberg:


Thrillberg:
I'm almost certain we're not going to want to go down this path. Or at least not this way.

PR #63, opened 2020-09-08
WebSockets

Thrillberg:


Thrillberg:
@jreut I've added a client side commit that I think uses your server side code correctly. Here's what it does:
1) On loading the Vue instance, client creates a JS WebSocket instance.
2) On receiving a message of type `setId`, the client will check if it already has an id stored in localStorage. If yes, it will send back an `updateId` message to the server. Finally, it will set the given id on localStorage.

One incidental question is about this output from the server:
server_go_—_imperial
The `onUpdateid` function returns `nil` so the Go output will always have `<nil>` written here when handling `updateId`. Is that the desired behavior?

PR #64, opened 2020-09-17
[WIP] Create a rondel Vue component that renders SVG

sobelk:
This PR lays out a pattern in which components render SVG. Ultimately, direct DOM manipulation might be avoided if all of the games graphics—including the rondel, board, and flags—can be built similarly. 

For now, the rondel has been implemented as a Vue component called `rondel`. Click events are handled by a `rondel-slot` component. When clicked, the correct action is found in the list of available actions and dispatched to the Imperial engine.

n.b. https://html.spec.whatwg.org/#global-attributes


Thrillberg:
@sobelk I've merged #66 , which duplicates most of what's in here. Shall we close this PR?

PR #68, opened 2020-10-01
Componentize the map

Thrillberg:


Thrillberg:
Here's how it looks so far!!
Imperial

Thrillberg:
This is how it looks now and all the provinces are SVG files loaded through iteration as `<Province>` components.
Imperial

Thrillberg:
Got some flags to show! Too bad they're wonky and in the wrong places!
Imperial

Thrillberg:
Flags are looking alright!
Imperial

Thrillberg:
It's coming along! Just need Berlin and Russia.
Screenshot_2020-10-07 Imperial

Thrillberg:
And now we have Berlin and Russia. Map is DONE, I think!
image

PR #78, opened 2020-10-16
Fix browser reload issue

Thrillberg:


server/server.go
```diff
@@ -56,7 +58,12 @@ func NewPlayerId() (PlayerId, error) {
 	return PlayerId(builder.String()), nil
 }
 
-type Action interface{}
+type Action struct {
```
Thrillberg: This struct is unnecessary.

PR #98, opened 2020-12-05
server: sweep away some cookie crumbs

jreut:


server/server.go
```diff
@@ -219,15 +220,13 @@ func handleHealth(w http.ResponseWriter, r *http.Request) {
...
 
 func handleWebsocket(w http.ResponseWriter, r *http.Request) {
-	userId := getUserIdFromCookie(r)
-	responseHeader := setUserIdToCookie(w, userId)
-
-	if len(userId) == 0 {
-		userId = NewUserId()
-		responseHeader = setUserIdToCookie(w, userId)
+	userId := NewUserId()
+	if cookie, err := r.Cookie(cookiename); err == nil {
```
Thrillberg, line 224: Ooohh I think I didn't see `r.Cookie` in the docs. Is this somehow preferable to `r.Header.Get("Cookie")`?

server/server.go
```diff
@@ -219,15 +220,13 @@ func handleHealth(w http.ResponseWriter, r *http.Request) {
...
+		userId = UserId(cookie.Value)
 	}
+	http.SetCookie(w, mintCookie(userId))
```
Thrillberg, line 227: What is the difference between doing `http.SetCookie` and the way I was trying it (setting it on the `responseHeader`)?

server/server.go
```diff
@@ -219,15 +220,13 @@ func handleHealth(w http.ResponseWriter, r *http.Request) {
...
 
-	ws, err := upgrader.Upgrade(w, r, responseHeader)
+	ws, err := upgrader.Upgrade(w, r, w.Header())
```
Thrillberg: I believe the last argument here can be `nil`, since we haven't done anything with `w`. Is that right?

server/server.go
```diff
@@ -251,24 +250,18 @@ func handleWebsocket(w http.ResponseWriter, r *http.Request) {
...
-	cookie := r.Header.Get("Cookie")
-	parsedCookie, cookieErr := url.ParseQuery(cookie)
-	if cookieErr != nil {
-		log.Println("Cookie", cookieErr)
+func mintCookie(userId UserId) *http.Cookie {
+	c := &http.Cookie{
+		Name:     cookiename,
+		Value:    string(userId),
+		MaxAge:   86400,
+		SameSite: http.SameSiteLaxMode,
```
Thrillberg, line 258: This makes way more sense than the mysterious `2` I had in there before. Thanks for making the code so much clearer!

Thrillberg (approved):
Looks great to me! Do you know if this technically "fixes" anything vs improving code readability/using more normal patterns? (Not to demean the latter, just wondering if I should expect better functioning when we deploy this!)

Thanks so much for doing this!

server/server.go
jreut: Yep! `http.Request.Cookie` parses the header and extracts the appropriate cookie by name. The [syntax of the Cookie header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) is a little different from a URL query string, so the previous version using `net/url.ParseQuery` won't work if the browser ever sends more than one cookie.

server/server.go
jreut: We have indeed done something with `w`! We have mutated it by adding a "Set-Cookie" header. I think you're right, though. A quick test confirms the upgrader sends the appropriate "Set-Cookie" header when the third argument is `nil`. I'll fix that.

server/server.go
jreut: `http.SetCookie` is a sort of dual to `http.Request.Cookie` in that the former "unparses" cookies into their appropriate "Set-Cookie" header value and the latter parses "Cookie" header values into their appropriate `*http.Cookie` representation. This refactoring doesn't change any behavior; indeed we were already using `http.SetCookie(w, cookie)`. This change just omits some unnecessary dancing around with `w.Header()`

Thrillberg:
Thanks for explaining everything to me! Merge when ready!

PR #99, opened 2020-12-05
server: add coarse locking to global state

jreut:


server/server.go
```diff
@@ -9,14 +9,28 @@ import (
...
-var users = map[UserId]UserName{}
-var games = map[GameId]*Game{}
+type globalState struct {
+	*sync.Mutex
+	connections map[UserId]*Conn
+	users       map[UserId]UserName
+	games       map[GameId]*Game
+}
+
+// state is this server's global state. It is coarsely locked by
```
Thrillberg, line 25: "Coarsely" here indicates that we lock all of state at a time and not one attribute at a time, yes?

server/server.go
```diff
@@ -132,12 +146,14 @@ func (c *Conn) write(v interface{}) error {
 
 // UpdateState sends registered users and open games to the client.
 func (c *Conn) UpdateState(id UserId) error {
-	if err := c.UserRegistered(users); err != nil {
-		log.Println(users, "UserRegistered")
+	state.Lock()
+	defer state.Unlock()
```
Thrillberg, line 150: Why do this vs putting `state.Unlock()` at the end of the function? It seems safer for future edits to the function so future you or me doesn't need to remember to keep moving it down. Are there other reasons or just that?

server/server.go
```diff
@@ -132,12 +146,14 @@ func (c *Conn) write(v interface{}) error {
 
 // UpdateState sends registered users and open games to the client.
 func (c *Conn) UpdateState(id UserId) error {
-	if err := c.UserRegistered(users); err != nil {
-		log.Println(users, "UserRegistered")
+	state.Lock()
```
Thrillberg, line 149: Just by virtue of having a `Mutex` on `state`, we're able to call `state.Lock()`? That seems wild to me. (Like, we wouldn't call `state.Mutex.Lock()`? That's more what I'd expect.)

Thrillberg (approved):
This looks awesome-snausage to me!! I'm wondering if a future elaboration could be to set up `state` such that it embeds its own logic around when to do a lock or not. Like, if we put something in the way of the elements right on the map, so we'd call `state.addUser(user: user)` and then `addUser` is one (of a few) functions that includes the two lines around locking and unlocking. Does that make any sense? I'm just trying to think of ways of pushing down the locking so it's not as visible at the level of code we're likely to spend more time in.

Same question as the other PR: will I see performance changes/improvements from these changes? Either way, thanks so much for helping with this!!

server/server.go
jreut: Yes indeed, that's exactly what I mean!

server/server.go
jreut: Yerp, this is a perhaps obscure [language feature](https://golang.org/ref/spec#Struct_types).

> A field declared with a type but no explicit field name is called an embedded field.
> ...
> A field or method f of an embedded field in a struct x is called promoted if x.f is a legal selector that denotes that field or method f.
>
> Promoted fields act like ordinary fields of a struct except that they cannot be used as field names in composite literals of the struct.

You _can_ also say `state.Mutex.Lock()`; the way I've written it here is just a shorthand.

server/server.go
jreut: It's actually safer to use a `defer`. What if we `panic()` between calling `state.Lock()` and `state.Unlock()`? Every call to [`sync/Mutex.Lock()`](https://golang.org/pkg/sync/#Mutex.Lock) simply waits if the mutex is already locked. If one goroutine fails without unlocking the mutex, it will remain locked forever. `defer`, on the other hand, lets us schedule unlocking for the end of the function. Deferred statements also run during panic recovery, which is the behavior we want.

Although Go doesn't use the term, Go panics are sort of like Ruby exceptions and Go defers are spiritually similar to Ruby try/catch blocks.

jreut (commented):
> 'm wondering if a future elaboration could be to set up `state` such that it embeds its own logic around when to do a lock or not.

That sounds like a good idea to me. It's definitely tedious and error-prone to need to remember about the mutex every time you want to access global state.

> Same question as the other PR: will I see performance changes/improvements from these changes? Either way, thanks so much for helping with this!!

We'll most likely see a slight performance degradation here, since each request that touches global state will need to wait in line behind any other requests that call `state.Lock()` before them. I don't think that's much of a problem now, though. The operations we're performing in the so-called "critical sections" (the code between locking and unlocking) are likely fast enough and our performance envelope lax enough that we won't notice.

Thrillberg:
Thanks for explaining everything to me! Merge when ready!

PR #101, opened 2020-12-12
Channels in the server

Thrillberg:


Thrillberg:
@jreut if you get a chance could you review this PR? I've converted all the registered handlers into our shiny new channel way. I haven't changed the messages themselves, though, figuring that we could do that with a follow up to this.

PR #107, opened 2020-12-26
demo: use database/sql with pgx

jreut:


Thrillberg:
@jreut This is great! I (rather painlessly) made it work for users to have `id` and `name` fields. And it works! Here are my thoughts on what to do next:
1) Figure out implications for publishing these changes to AWS. I presume I'll need to hook up a production database there.
2) Also, I notice that the database this is using locally is my `eric.tillberg` (username) db. I'd rather have it have its own `imperial` postgres database.
3) There's bugginess around if a user registers with a "taken" name. I think it breaks for both players so that's a problem.
4) I think if we forego passwords then auth is pretty much done so... onward to a `games` table! (?)

Issue #127, opened 2021-01-19
Use channels properly


Thrillberg:
Currently, `app/javascript/src/router/index.js` routes everything to `AppearanceChannel`, which isn't correct. We probably want the following other channels:
- `JoinGameChannel`
- `OpenGameChannel`
- `GetGameLogChannel`
- `TickChannel`

If we do this, we can probably also simplify the data that is sent and dispense with the `kind` fields.

PR #139, opened 2021-01-20
Fractures one channel into separate channels

armahillo:
Ref #127


app/channels/appearance_channel.rb
```diff
@@ -1,83 +1,41 @@
-class AppearanceChannel < ApplicationCable::Channel
+class AppearanceChannel < ApplicationChannel
```
Thrillberg: I wonder if we can keep the inheritance the same (`ApplicationCable::Channel`) but put the contents of `ApplicationChannel` into `ApplicationCable::Channel`. Currently that model is empty and it seems like that may be a better place for the shared logic. What do you think?

app/channels/appearance_channel.rb
```diff
@@ -1,83 +1,41 @@
...
-    }
-    ActionCable.server.broadcast("appearance_channel", payload)
+    broadcast_users "appearance_channel", "updateUsers"
+    broadcast_games "appearance_channel", "updateGames"
   end
 
   def receive(data)
     case data["kind"]
     when "openGame"
+      # TODO: can this be userId also, so that it's the same as below?
```
Thrillberg: I would be fine with renaming `host` but this method and `joinGame` would still differ because `openGame` creates a new game. Does that make it worthwhile to change from `host` to `userId`? I'm not sure what all the implications would be!

app/channels/appearance_channel.rb
```diff
@@ -1,83 +1,41 @@
...
-      payload = {
-        kind: "updateGames",
-        data: {
-          games: Game.all.map do |game|
-            to_json(game)
-          end
-        }
-      }
-      ActionCable.server.broadcast("appearance_channel", payload)
+      broadcast_games "appearance_channel", "updateGames"
```
Thrillberg, line 13: I love how we're using these shared methods. Nice refactor!

app/channels/appearance_channel.rb
```diff
@@ -1,83 +1,41 @@
...
-  def to_json(game)
-    {
-      name: game.name,
-      id: game.id,
-      host: game.host.name,
-      players: game.users.map(&:name),
-      log: game.actions.map(&:data)
-    }
+  def game_from_data(data)
+    Game.includes(:actions).find(data["data"]["gameId"])
```
Thrillberg, line 37: I'm curious about the `includes(:actions)` here. Is it necessary? What is it doing?

Thrillberg (approved):

PR #152, opened 2021-01-22
Add devise 114

Thrillberg:


Thrillberg:
@armahillo A note on nomenclature:

**identified** is when a user has submitted a username but no email and password
**registered** is when they have a username, an email and a password

What do you think of those words?

Here's what I think is left to do on this:

~~- [ ] Allow people to log out (on second thought, I think this isn't necessary for right now. I think it should go along with being able to log in)~~
- [x] When user is not yet **identified**, do not let them submit email and password or start a game
- [x] There is a bug when trying to identify yourself, the route is incorrectly pointing to `POST` on `/user` rather than `/users`

Before merging:
- [x] Resolve console errors
- [x] Linter
- [ ] Tests?

PR #157, opened 2021-01-23
Complete TODO tests

Thrillberg:


Thrillberg:
Still need to update the front end to accommodate forcing Investor and skipping force Investor.

Thrillberg:
This is very buggy and it's hard to tell why. Let's resolve #159 first.

PR #167, opened 2021-01-25
Add admin area for granting and taking bonds

### LINKED ISSUE #166: Create admin mode
~~It would be nice, especially for #129 and #130, if we could create an admin mode where I can arbitrarily add and remove bonds from players. Specifically, I want to be able to deprive a player of all their bonds so that they acquire a Swiss Bank. Turns out that's hard/tedious to do in the normal course of play.~~

~~On second thought, I think we need a way to input test representing a game log that can then be saved to the database as a new game. Then from there we can play around with it.~~

On third thought, I'm not sure how we're going to accomplish this. Translating from JS to text back into something ingested by Ruby is actually quite tricky and I'm not sure about a cleaner way to do it.

Thrillberg:
Fixes #166

### DIFF
--- a/app/javascript/src/components/Bond.vue
+++ b/app/javascript/src/components/Bond.vue
@@ -2,6 +2,7 @@
   <div
     class="m-2 border border-gray-500 p-1"
     v-if="nation === bond.nation.value"
+    @click="click_handler(bond.number, bond.nation)"
   >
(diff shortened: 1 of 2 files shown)


Thrillberg:
Urgh, I think we can't merge this because it's not quite working as intended. The log keeps interfering, of course, and fixing the admin interventions. I think this won't actually help with debugging.

Issue #134, opened 2021-01-19
Show all provinces that are valid plays


Thrillberg:
When maneuvering or importing we should be able to see all available provinces to manipulate. Highlight them all and come up with a "brighter" or different highlight for the selected one.

PR #179, opened 2021-01-29
Improve the visibility of valid options in the game

dmcarmo:
Current solution for issue #134 is to have an _hoverable_ class that looks different from the default style and also the _hoverable:hover_ config.
This will highlight not only the valid provinces to choose for an action but also the destination provinces for armies and the valid slots in the rondel.


Thrillberg (approved):
This is a vast improvement! Thank you!!

PR #193, opened 2021-01-31
Improve game log text and structure

dmcarmo:
Implements multiple methods for humanizing the game logs

### DIFF
--- a/app/javascript/src/components/GameLog.vue
+++ b/app/javascript/src/components/GameLog.vue
@@ -1,0 +1,38 @@
+<template>
+  <div class="border border-black rounded p-2 m-2">
+    
+    <div>
+      <GameLogEntry
+        v-for="(entry, index) in processedLog()"
+          :events="entry"
+          :key="index"
+      ></GameLogEntry>
+    </div>
+  </div>
+</template>
+
+<script>
+import GameLogEntry from "./GameLogEntry.vue";
+export default {
+  name: "GameLog",
+  props: { log: Array },
+  components: { GameLogEntry },
+  methods: {
+    processedLog() {
+      let rawlog = this.log;
+      let entries = [];
+      rawlog.forEach((action) => {
+        if (action.type === "initialize") {
+          entries.push({event: action});
+        } else if (action.type === "rondel") {
+          entries.push({event: [action]});
+        }
+        else {
+          entries[entries.length - 1].event.push(action);
+        }
+      });
+      return entries.slice().reverse();
+    } 
+  }
+}
+</script>
--- a/app/javascript/src/components/GameLog.vue
+++ b/app/javascript/src/components/GameLog.vue
@@ -1,13 +1,10 @@
 <template>
   <div class="border border-black rounded p-2 m-2">
-    
-    <div>
-      <GameLogEntry
-        v-for="(entry, index) in processedLog()"
-          :events="entry"
-          :key="index"
-      ></GameLogEntry>
-    </div>
+    <GameLogEntry
+      v-for="(entry, index) in processedLog()"
+        :events="entry"
+        :key="index"
+    ></GameLogEntry>
   </div>
 </template>
 
@@ -22,16 +19,14 @@ export default {
       let rawlog = this.log || [];
       let entries = [];
       rawlog.forEach((action) => {
-        if (action.type === "initialize") {
-          entries.push({event: action});
-        } else if (action.type === "rondel") {
+        if (action.type === "initialize" || action.type === "rondel") {
           entries.push({event: [action]});
         }
         else {
           entries[entries.length - 1].event.push(action);
         }
       });
-      return entries.slice().reverse();
+      return entries.reverse();
     } 
   }
 }
--- a/app/javascript/src/components/GameLog.vue
+++ b/app/javascript/src/components/GameLog.vue
@@ -19,7 +19,7 @@ export default {
   components: { GameLogEntry },
   methods: {
     processedLog() {
-      let rawlog = this.log;
+      let rawlog = this.log || [];
       let entries = [];
       rawlog.forEach((action) => {
         if (action.type === "initialize") {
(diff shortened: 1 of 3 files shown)


app/javascript/src/components/GameLog.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div class="border border-black rounded p-2 m-2">
+    
+    <div>
```
Thrillberg: Is this `div` necessary? Can we remove it?

app/javascript/src/components/GameLog.vue
```diff
@@ -0,0 +1,38 @@
...
+  components: { GameLogEntry },
+  methods: {
+    processedLog() {
+      let rawlog = this.log || [];
+      let entries = [];
+      rawlog.forEach((action) => {
+        if (action.type === "initialize") {
+          entries.push({event: action});
+        } else if (action.type === "rondel") {
+          entries.push({event: [action]});
```
Thrillberg, line 23: I wonder if this can be tightened up like this:
```javascript
if (action.type === "initialize" || action.type === "rondel") {
  entries.push({event: [action]});
}
```
This would be a substantive change in that the `action` for "initialize" would be a single element in an array rather than directly the element itself. But I feel like that will keep the pattern a little tighter. Would my suggested change break things?

app/javascript/src/components/GameLog.vue
```diff
@@ -0,0 +1,38 @@
...
+        }
+        else {
+          entries[entries.length - 1].event.push(action);
+        }
+      });
+      return entries.slice().reverse();
```
Thrillberg: Very small point but I don't _think_ you need `slice()` here. `reverse()` does indeed mutate the array but since we re-generate the array each time this is called, I think that it would be harmless to make the mutation here. Does it work if we remove `slice()`?

app/javascript/src/components/GameLogEntry.vue
```diff
@@ -1,17 +1,132 @@
 <template>
-  <div class="border border-black rounded p-2 m-2">
+  <div v-if="Array.isArray(event)" class="border border-black rounded p-2 m-2 rondel">
```
Thrillberg: Ah, my recommendation above would break this logic. I think it's worth investigating this a small bit further. Can we turn our conditional on something other than the array-ness of `event`? Perhaps `event.type` would be a more direct way to get the logic to work?

app/javascript/src/components/GameLogEntry.vue
```diff
@@ -1,17 +1,132 @@
...
+    <div v-for="(action, index) in event" :key="(index + Math.random()).toString()">
```
Thrillberg: I'd like to get `Math.random()` out of here if possible. Don't worry about the duplicate index error with `GB0`, I've resolved that elsewhere in `Province` (and it's entirely unrelated to your changes here). Can we make the key just be `index`?

app/javascript/src/components/GameLogEntry.vue
```diff
@@ -1,17 +1,132 @@
...
+      <div v-if="action.type === 'rondel'">
+        <div>
+          <svg 
+            class="inline-block mr-1" 
+            xmlns="http://www.w3.org/2000/svg" 
+            width="30" 
+            height="20"
+            >
+            <Flag :nation="action.payload.nation.value" :width="(30).toString()"></Flag>
```
Thrillberg: I know this is a (bad) pattern I have elsewhere in the codebase but you can do:
```html
width="30"
```
Without the `:`, you don't need to `toString()` the 30.

app/javascript/src/components/GameLogEntry.vue
```diff
@@ -1,17 +1,132 @@
...
+      return (word[0].toUpperCase() + word.substring(1));
+      } else {
+        return "";
+      }
+    },
+    processAction(action) {
+      let notImplemented = "NOT IMPLEMENTED";
+      switch (action.type) {
+        case "initialize":
+          return notImplemented;
```
Thrillberg, line 59: Can we implement this now or would you rather wait until later. Either way is totally ok with me!

Thrillberg (commented):
This is really great work, David! Thanks again for (stubbornly) sticking with it. I have some comments inside, which you're welcome to address if you want. But if you want a break from this, I'm also happy to carry this puppy home. Let me know what you prefer!

app/javascript/src/components/GameLog.vue
dmcarmo: Nope, it can be removed.

app/javascript/src/components/GameLog.vue
dmcarmo: I did some tests and it looks like we can remove `slice()`.

app/javascript/src/components/GameLogEntry.vue
dmcarmo: Yes, it looks like we can change it back to only `index`.

app/javascript/src/components/GameLogEntry.vue
dmcarmo: Ok, I still need to send it as a string so I'm changing it to `:width="'30'"`
**edit**
Sorry, I forgot to test without the `:` so what I said before doesn't make any sense. Yes, it works and I corrected it on my side.

app/javascript/src/components/GameLogEntry.vue
dmcarmo: I'll implement this now.

app/javascript/src/components/GameLog.vue
dmcarmo: Yes, it would need some other changes for this to work but it's worth it.

app/javascript/src/components/GameLogEntry.vue
dmcarmo: As mentioned before it's worth changing this. :+1:

Thrillberg (approved):
Looks great! Please merge whenever you want to! 🎉

Issue #211, opened 2021-02-05
Disallow others from playing in my solo mode game


Thrillberg:
Currently, anyone can submit moves in a solo mode game.

PR #214, opened 2021-02-06
Fix for issue #211 - Disallow others from playing in my solo mode game

dmcarmo:
Instead of only checking if soloMode is true it also checks is the profile username is in the current game's list of players.

### DIFF
--- a/app/javascript/src/components/GameDetails.vue
+++ b/app/javascript/src/components/GameDetails.vue
@@ -47,7 +47,7 @@
       ></Rondel>
     </div>
     <div
-      v-if="game.importing && (profile.username === controllingPlayerName || game.soloMode)"
+      v-if="game.importing && (profile.username === controllingPlayerName || (game.soloMode && profile.username in game.players))"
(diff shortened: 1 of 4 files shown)


app/javascript/src/components/GameDetails.vue
```diff
@@ -47,7 +47,7 @@
       ></Rondel>
     </div>
     <div
-      v-if="game.importing && (profile.username === controllingPlayerName || game.soloMode)"
+      v-if="game.importing && (profile.username === controllingPlayerName || (game.soloMode && profile.username in game.players))"
```
Thrillberg, line 50: Can this be written as `game.players.includes(profile.username)`? I think that's more idiomatic since the `in` word is generally used for iterating through the collection, if I understand correctly.

Thrillberg (commented):
This is great! I left one comment, which is applicable to most of the changes in this PR. Once that is resolved, it'll be ready to go!

Thrillberg (approved):
Nevermind, 👍!!

PR #229, opened 2021-02-09
Fixes bin/update, which was referencing npm instead of yarn

armahillo:


Thrillberg:
Is there a reason npm wasn't working? @dmcarmo do I recall correctly that you had an issue with yarn or something along the way? I think our goal was to avoid yarn but I don't remember why!

dmcarmo:
> Is there a reason npm wasn't working? @dmcarmo do I recall correctly that you had an issue with yarn or something along the way? I think our goal was to avoid yarn but I don't remember why!

I checked my history and I did try `yarn install` first which gave the following error:
`error @rails/webpacker@6.0.0-beta.4: The engine "node" is incompatible with this module. Expected version ">=10.22.1 || ^12 || >=14". Got "10.19.0"
error Found incompatible module.`

I then ran `npm install` and everything worked fine.

PR #340, opened 2021-02-28
Dockerize application

joleal:
This change includes a dockerfile and a docker-compose.yml which dockerizes the environment. I don't think the webpacker is working correctly so there's still a bit of work to be done there, but the rest should be fine.


Thrillberg (approved):
I love it! Thanks so much for dockerizing this, João! Feel free to merge when you want to. 🎉

PR #398, opened 2021-03-29
Handle 3-way fight

Thrillberg:


Thrillberg:
Don't merge until verklempt night is over. This will introduce breaking changes.

Issue #384, opened 2021-03-26
Limit total units each country can have


Thrillberg:
AH may not have more than 10 armies and 6 fleets
GB may not have  more than 6 armies and 10 fleets
all other nations cannot have more than 8 armies and 8 fleets

PR #400, opened 2021-03-29
Implements unit limits for import and production

dmcarmo:
Fix for #384 
Adds unit limits for each nation in standardSetup.js
When importing, if the unit limit for that nation is crossed the action is reset.
When producing, if the unit limit for that nation is reached no more units of that type are produced.

### DIFF
--- a/app/javascript/lib/imperial.js
+++ b/app/javascript/lib/imperial.js
@@ -191,6 +191,7 @@ export default class Imperial {
     this.order = s.order;
     this.players = s.players;
     this.provinces = s.provinces;
+    this.unitLimits = s.unitLimits;
     this.units = this.initializeUnits(s.units);
     this.currentPlayerName = this.getStartingPlayer();
     this.previousPlayerName = this.currentPlayerName;
@@ -1038,9 +1039,13 @@ export default class Imperial {
           .forEach(province => {
             const units = this.units.get(nation).get(province);
             if (this.nobodyIsOccupying(province, nation)) {
-              if (this.provinces.get(province).factory === "shipyard") {
+              if (this.provinces.get(province).factory === "shipyard" &&
+                  !this.overUnitLimit(action.payload.nation, "fleets", 1)
+              ) {
                 units.fleets++;
-              } else {
+              } else if (this.provinces.get(province).factory != "shipyard" &&
+                        !this.overUnitLimit(action.payload.nation, "armies", 1)
+              ) {
                 units.armies++;
               }
             }
@@ -1214,7 +1219,7 @@ export default class Imperial {
     const unoccupiedHomeProvinces = [...homeProvinces].filter(province => {
       let unoccupied = true;
       for (const [nation] of this.nations) {
-        if (this.units.get(nation).get(province).armies > 0) {
+        if (nation != action.payload.nation && this.units.get(nation).get(province).armies > 0) {
           unoccupied = false
         }
       }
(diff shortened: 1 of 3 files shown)


app/javascript/lib/standardSetup.js
```diff
@@ -182,8 +182,21 @@ export default ({ players, provinceNames }) => {
     }
     provinces.set(province, { factory });
   }
+
+  const unitLimits = new Map();
+
+  for (const nation of Nation) {
+    if (nation == Nation.AH) {
+      unitLimits.set(nation, {armies: 10, fleets: 6});
+    } else if (nation == Nation.GB) {
```
Thrillberg: It shouldn't make much difference but for the sake of regularity, can we use `===` here? And on line 189?

app/javascript/lib/imperial.js
```diff
@@ -1214,7 +1219,7 @@ export default class Imperial {
     const unoccupiedHomeProvinces = [...homeProvinces].filter(province => {
       let unoccupied = true;
       for (const [nation] of this.nations) {
-        if (this.units.get(nation).get(province).armies > 0) {
+        if (nation != action.payload.nation && this.units.get(nation).get(province).armies > 0) {
```
Thrillberg: Can we change this to `!==`?

app/javascript/lib/imperial.js
```diff
@@ -1226,12 +1231,33 @@ export default class Imperial {
...
+      this.availableActions = availableActions;
+      this.importing = true;
+      return;
+    }
+
+    if (this.fleetCount(action.payload.nation) == this.unitLimits.get(action.payload.nation).fleets) {
+      this.availableActions = availableActions;
+      this.importing = true;
+      return;
+    }
```
Thrillberg: I'm thinking this and the above section about armies can probably be deleted now. Do tests pass if we delete them? They should, I think, because the logic below should filter out any additional import actions.

app/javascript/lib/imperial.js
```diff
@@ -1226,12 +1231,33 @@ export default class Imperial {
...
+
+    // const imports = Array.from({length: this.maxImports}, (_, i) => i + 1);
+    // this.unitLimits.get(action.payload.nation)[unitType]);
+    // overUnitLimit(nation, unitType, numberOfPlacements)
```
Thrillberg: Let's remove the comments here.

app/javascript/lib/imperial.js
```diff
@@ -1555,14 +1607,33 @@ export default class Imperial {
...
+  fleetCount(nation) {
+    let out = 0;
+    for (const [, units] of this.units.get(nation)) {
       out += units.fleets;
     }
     return out;
   }
 
+  overUnitLimit(nation, unitType, numberOfPlacements) {
+    if (unitType === "armies") {
```
Thrillberg: For uniformity, what do you think of using `"army"` instead of `"armies"`? It's a very very minor thing but I think generally elsewhere in the codebase it would be `"army"`.

app/javascript/lib/imperial.test.js
```diff
@@ -749,6 +865,454 @@ describe("imperial", () => {
...
+          game.tick(
+            Action.rondel({ slot: "import", cost: 0, nation: Nation.AH })
+          );
+
+          expect(game.availableActions).toEqual(availableActions);
+          expect(game.nations.get(Nation.AH).rondelPosition).toEqual("import");
+          expect(game.importing).toEqual(true);
+        });
+
+        test("importing when armies and fleets are both at the limit", () => {
```
Thrillberg, line 1265: Good idea to test this! 👍

app/javascript/lib/imperial.test.js
```diff
@@ -903,7 +1467,7 @@ describe("imperial", () => {
         });
       });
 
-      describe("production1 or production2", () => {
+      describe.only("production1 or production2", () => {
```
Thrillberg: We should remove the `.only` from here.

app/javascript/lib/imperial.test.js
```diff
@@ -938,7 +1502,7 @@ describe("imperial", () => {
               expect(game.units.get(Nation.AH).get("a").armies).toEqual(0);
             });
 
-            test("a unit is not produce in a province that has an occupied factory", () => {
+            test("a unit is not produced in a province that has an occupied factory", () => {
```
Thrillberg, line 1526: Thanks for fixing this typo!! 😬

Thrillberg (commented):
This looks great to me! Thank you so much for all your work you put into this, David! I left a couple comments so once those are addressed, it's probably ready to merge.

app/javascript/lib/imperial.js
dmcarmo: Good point! Actually one of the tests failed when i commented this block of code. Found out that the expected results for that test weren't correct and fixed that.

Thrillberg (approved):
Love it! Good to merge!

PR #682, opened 2023-03-17
Added minor instruction steps to README

GitGubM8:
## Comments:
- Probably unnecessary information for more experienced webdevs, but it'd be a nice-to-have for someone like me who's less familiar with web techs
- Obviously feel free to reword/correct any information that's unclear or straight-up wrong

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,15 @@ git clone https://github.com/Thrillberg/imperial.git
 ```
 cd imperial
 ```
-3. Run the Rails setup command:
+3. Begin the PostgreSQL service:
+```
+sudo service postgresql start
+```
+4. In a separate terminal, start a Redis server:
+```
+redis-server
+```
+5. Run the Rails setup command:
 ```
 bin/setup
 ```
@@ -32,17 +40,22 @@ bin/setup
 ```
 sudo su postgres -c "psql rails_server_development -c 'CREATE EXTENSION pgcrypto;'"
 ```
-4. Run the npm install command:
+- If you get the error `‘ruby\r’: No such file or directory`, make sure your editor's line-endings are set to `LF` instead of `CRLF`.
+6. Run the npm install command:
 ```
 npm install
 ```
-5. In one terminal window, run webpack:
+7. In one terminal window, run webpack:
 ```
 bin/webpack --watch
 ```
-6. Keep webpack running and in a separate terminal window, run the Rails server:
+8. Keep webpack running and in a separate terminal window, run the Rails server:
 ```
 rails s
+```
+9. You can now view the app at
+```
+localhost:3000/
 ```
 If any of the above instructions did not work for you, please do not hesitate to [open an issue](https://github.com/Thrillberg/imperial/issues/new).
 
--- a/README.md
+++ b/README.md
@@ -25,10 +25,20 @@ git clone https://github.com/Thrillberg/imperial.git
 cd imperial
 ```
 3. Begin the PostgreSQL service:
+- If instaleld through `brew`
+```
+brew services postgresql start
+```
+- If installed through `apt`
 ```
 sudo service postgresql start
 ```
-4. In a separate terminal, start a Redis server:
+4. Start a Redis server:
+- If installed through `brew`,
+```
+brew services start redis
+```
+- If installed through `apt`, in a separate terminal
 ```
 redis-server
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,26 +24,39 @@ git clone https://github.com/Thrillberg/imperial.git
 ```
 cd imperial
 ```
-3. Run the Rails setup command:
+3. Begin the PostgreSQL service:
+```
+sudo service postgresql start
```
Thrillberg, line 34: Does this command work? I would have thought it would be this (if pg is installed with homebrew):
```
brew services postgresql start
```
In any event, I'd like to avoid invocations of `sudo` here if possible.

README.md
```diff
@@ -24,26 +24,39 @@ git clone https://github.com/Thrillberg/imperial.git
...
+```
+4. In a separate terminal, start a Redis server:
+```
+redis-server
```
Thrillberg, line 43: Similarly, if redis is installed with homebrew then you can do:
```
brew services start redis
```
And not need a separate terminal open. I'd be happy to have such instructions included if they are helpful! 😄

README.md
```diff
@@ -24,26 +24,39 @@ git clone https://github.com/Thrillberg/imperial.git
...
+5. Run the Rails setup command:
 ```
 bin/setup
 ```
 - If you get the error `PG::InsufficientPrivilege: ERROR:  permission denied to create extension "pgcrypto"` run this command:
 ```
 sudo su postgres -c "psql rails_server_development -c 'CREATE EXTENSION pgcrypto;'"
 ```
-4. Run the npm install command:
+- If you get the error `‘ruby\r’: No such file or directory`, make sure your editor's line-endings are set to `LF` instead of `CRLF`.
```
Thrillberg, line 53: ooh I didn't know about this one. Great!

README.md
```diff
@@ -24,26 +24,39 @@ git clone https://github.com/Thrillberg/imperial.git
...
 bin/webpack --watch
 ```
-6. Keep webpack running and in a separate terminal window, run the Rails server:
+8. Keep webpack running and in a separate terminal window, run the Rails server:
 ```
 rails s
 ```
+9. You can now view the app at
+```
+localhost:3000/
```
Thrillberg, line 68: 👍 👍 👍

Thrillberg (commented):
Thank you for this PR, @GitGubM8! 🎉 I've got a couple clarifying questions in the review, but generally looks very helpful!

README.md
GitGubM8: Ah I guess I installed it directly using `apt install` instead of `brew` so this doesn't work for me, let me think a bit on how to best reword this so it suits everyone

Thrillberg: Oh that makes sense! If you want to specify that this line works where `apt install` is used rather than `brew`, that would be fine. I think `brew` is popular but obviously not the _only_ choice 😄 Out of curiosity, is the `sudo` necessary?

GitGubM8: Unfortunately `sudo` is necessary in the case where it's installed through `apt`

How does this look
[ef71461](https://github.com/Thrillberg/imperial/pull/682/commits/ef714611f3e1c8fdc8eec61b117aaee750dbad77)
[6424bc0](https://github.com/Thrillberg/imperial/pull/682/commits/6424bc0215b8a97c30b052d9d62f94c16851adfa)

Alternatively, Im happy to remove the `apt` lines too. I'd imagine some 90% of the ppl would've setup through `brew`. I was just bumbling my way through and installing through `brew` never occurred to me

README.md
GitGubM8: Same as the `PostgreSQL` comment

[ef71461](https://github.com/Thrillberg/imperial/pull/682/commits/ef714611f3e1c8fdc8eec61b117aaee750dbad77)
[6424bc0](https://github.com/Thrillberg/imperial/pull/682/commits/6424bc0215b8a97c30b052d9d62f94c16851adfa)

README.md
Thrillberg: Ok sounds good! Let's keep the `apt` lines in, just in case.

I made a mistake here! It should be `brew services start postgresql`. I was just going off memory before but I just confirmed that it works this way.

GitGubM8: Fixed in
[bd7aa08](https://github.com/Thrillberg/imperial/pull/682/commits/bd7aa08114e498b8cf0ba3a42e2d1e8011477293)

Issue #676, opened 2023-02-20
Add move numbers to game log


Thrillberg:
<img width="676" alt="Screenshot 2023-02-20 at 3 11 36 PM" src="https://user-images.githubusercontent.com/10481391/220191073-c4646cb0-1574-4978-95da-96c4a12595aa.png">

PR #687, opened 2023-03-18
676: Added Turn count to GameLogs

GitGubM8:
# Closes:
- https://github.com/Thrillberg/imperial/issues/676

## Screenshots:
image

## Credits:
- Credits to @Thrillberg for pair programming with me


GitGubM8:
Do note that this results in `undo` turns having their own turn number

image

But I think that's fine if not even better since it adds clarity when we want to reference the turn in a bug report

Thrillberg:
Looks great! Thanks again @GitGubM8!

Issue #684, opened 2023-03-17
Optimize neighborsFor function


Thrillberg:
This is recursive. And slow.

Thrillberg:
Wellp, I thought I fixed this but it ended up breaking games so I'm reverting my "fix". https://github.com/Thrillberg/imperial/commit/e97ad5142933b3fd33a771868480d0ca38f767b5

Thrillberg:
I think this is resolved here: https://github.com/Thrillberg/imperial/commit/db6db913fa9d0656bbb299171a52a4b4facb6aff

PR #710, opened 2023-03-23
Small pathsFrom refactor

Thrillberg:
Part of #684


GitGubM8 (approved):
Not familiar enough with this but looks good to me so far

I'd suggest delaying merging this a bit to see if any games break from the Factory_Finance PR, so we can more easily isolate issues

Thrillberg:
> I'd suggest delaying merging this a bit to see if any games break from the Factory_Finance PR, so we can more easily isolate issues

Totally agree! Do you think it's safe to merge this now/soon?

Issue #302, opened 2021-02-21
Allow players to finance the nations


dmcarmo:
When a nation doesn't have enough funds to build a factory or import units, the controlling player may spend his own funds to finance these actions.
But the rules also allow other players to finance nations that aren't controlled by them.

According to the rulebook:
"The players hold the treasuries they control in trust. They are not allowed to dispose of the treasuries personally nor lend or give away money from the treasury. However, they are allowed to give personal cash to the treasury at any time in order to support it financially."

And here's a clarification from the designer:
https://www.boardgamegeek.com/thread/302718/article/2269881#2269881

Thrillberg:
This will be interesting to implement. The concept of being able to give personal cash at **any time** strains our `availableActions` system where all actions are explicitly listed. How do we represent this in the game logic? Do we have a single action for each possible contribution to each possible nation? The multiplication gets wild with that...

Or maybe we implement something that programmatically allows any contribution actions up to a limit (the player's cash amount)? Then the mulitplication is only by 6, one for each nation.

I think I like the second option more but I don't understand as well how to implement it at the moment. I feel like it would break our [naive check for available actions](https://github.com/Thrillberg/imperial/blob/c9c3a668c921b5ab04c4a192b3a0380e5df44c40/app/javascript/lib/imperial.js#L53) at every tick.

dmcarmo:
I think we can split this into two goals:
1. the active player is able to finance the current nation if needed
1. anyone can finance a nation at any time (which would need some sort of channel for the players to negotiate this)

The first one is probably easier and more urgent to implement.
The second one is trickier as it also requires negotiation/communication between players.

Thrillberg:
I'm not sure I would divide it that way, personally. For one thing, is negotiation or communication necessary? I think to start we can just have players be able to contribute and presume any in-game communication is happening outside of the app (phone/Zoom call/Google Meet/Discord/telepathy).

For the mechanics of registering a "finance"-type action, I think it might be best if we make our check for available actions on each tick, linked in my previous comment, handle contributions too and not just disqualify a valid contribution because it is not an "available action". This will make our validation code a little more complicated but I think that's ok.

dmcarmo:
Sure, players can use any existing channel to communicate between themselves, but I'm not sure if you're obligated to accept any proposal so some sort of negotiation is needed.
I thought of breaking the problem in two because the first one is easier to solve, since it only involves the active player and the current nation.
[...]

PR #711, opened 2023-03-24
Factory finance

GitGubM8:
# Partially Closes:
- https://github.com/Thrillberg/imperial/issues/302

## Screenshots:
image

## Comments:
- Backwards compatible


app/javascript/lib/Entities/Board/Province.js
```diff
@@ -0,0 +1,58 @@
+import Entity from "../Entity.js"
+
+export default class Province extends Entity {
```
GitGubM8, line 3: Only a partial refactor, not meant to be the complete model just yet

Currently the `Province` entity is only created for evaluation purposes and immediately discarded afterward.

Once enough areas have been updated to support the new system, the goal is to flip all older designs into the newer one, and we'll no longer need to discard these entities after use

app/javascript/lib/imperial.js
```diff
@@ -1336,31 +1356,46 @@ export default class Imperial {
         return;
       }
       case 'factory': {
-        // If nation cannot afford to build a factory
-        if (this.nations.get(this.currentNation).treasury < 5) {
-          if (this.variant === 'withoutInvestorCard') {
```
GitGubM8, line 1341: Players can now intentionally waste the nation's time not building a factory

Slightly useful if you know your hold over a nation is temporary and really want to screw them for 1 whole round before you lose grasp

app/javascript/lib/action.js
```diff
@@ -64,7 +66,7 @@ export default {
   initialize: makeAction("initialize", ["players", "soloMode", "variant", "baseGame"]),
   blockCanal: makeAction("blockCanal", []),
   bondPurchase: makeAction("bondPurchase", ["nation", "player", "cost", "tradeInValue"]),
-  buildFactory: makeAction("buildFactory", ["province"]),
+  buildFactory: makeAction("buildFactory", ["nation", "province", "player", "nationCosts", "playerCosts"]),
```
Thrillberg: Do we need to specify `nation` here? Aren't all provinces hardcoded to which nation they belong (if any)?

app/javascript/lib/imperial.js
```diff
@@ -2399,6 +2373,11 @@ export default class Imperial {
         );
       }
 
+      if (action1.type === 'buildFactory' && action2.type === 'buildFactory') {
+        // accept migrations from older designs, which previously did not allow for funding nations direction
+        return action1.payload.province === action2.payload.province;
```
GitGubM8, line 2371: Backwards compatibility here

app/javascript/lib/action.js
```diff
@@ -81,6 +83,7 @@ export default {
   maneuver: makeAction("maneuver", ["origin", "destination"]),
   production: makeAction("production", ["province"]),
   rondel: makeAction("rondel", ["nation", "cost", "slot"]),
+  couldNotBuildFactory: makeAction("couldNotBuildFactory", ["nation"]),
```
Thrillberg, line 86: Is this action being used yet?

app/javascript/lib/action.js
GitGubM8: Ah yeah you're right, we can get `nation` through `graph.get(province).nation`, can't we?
Will fix

GitGubM8: Not quite sure how to get `game.board.graph` from within `GameLogEntry.vue`, going to leave it for now

Tangentially related, [e2d3726](https://github.com/Thrillberg/imperial/pull/711/commits/e2d372612c745d4855d5dc00635b368baad2e16a)

GitGubM8: Actually, I'm blind

Fixed here [06a08f6](https://github.com/Thrillberg/imperial/pull/711/commits/06a08f6ea787c95d59de0cb126cfac2ce7479ea4)

app/javascript/lib/Entities/Entity.js
```diff
@@ -0,0 +1,11 @@
+export default class Entity {
+    #representation;
```
Thrillberg: I'm not familiar with this in JS. What does `#representation` do here?

GitGubM8: Apparently, that's how we declare private modifiers in js

Helps ensure that `representation` is read-only since it's exposed through the getter but has no setter

app/javascript/lib/imperial.js
```diff
@@ -228,7 +241,7 @@ export default class Imperial {
       }
       case 'rondel': {
         this.previousPlayerName = this.currentPlayerName;
-        this.rondel(action);
+        this.rondelExecute(action);
```
Thrillberg: I wonder if a name like `advanceOnRondel(action)` is more descriptive? I totally agree that `rondel(action)` isn't as helpful as it could be!

app/javascript/lib/imperial.js
```diff
@@ -180,6 +190,9 @@ export default class Imperial {
         this.buildFactory(action);
         return;
       }
+      case 'couldNotBuildFactory':
```
Thrillberg: I don't _think_ we need this action type. If a factory cannot be built (nation AND player combined don't have enough money) then we can make the logic proceed without an extra action. We should keep the `annotatedLog` entry elsewhere in this PR, though, as it's useful to communicate to the players.

app/javascript/lib/imperial.js
```diff
@@ -894,10 +907,18 @@ export default class Imperial {
   }
 
   buildFactory(action) {
-    this.provinces.get(action.payload.province).factory = this.board.graph.get(
-      action.payload.province,
-    ).factoryType;
-    this.nations.get(this.currentNation).treasury -= 5;
+    const currentNation = this.nations.get(action.payload.nation ? action.payload.nation : this.currentNation);
```
Thrillberg: This seems like it potentially opens a vulnerability where someone could submit an action that asserts a nation is the current nation even though, according to the game logic, that nation isn't the current nation. I'd prefer to rely on `this.currentNation`, if possible.

app/javascript/lib/imperial.js
```diff
@@ -894,10 +907,18 @@ export default class Imperial {
...
+    const currentPlayer = this.players[action.payload.player || this.currentPlayerName];
```
Thrillberg: Same concern as above.

app/javascript/lib/Entities/Rondel.js
```diff
@@ -0,0 +1,86 @@
...
+
+  slotClockwiseTo(slot, slotCounts) {
+    let slotIndex = this.slotOrder.indexOf(slot);
+    slotIndex += slotCounts;
+
+    slotIndex %= this.slotOrder.length;
+    return this.slotOrder[slotIndex];
+  }
+
+  passedInvestor(exclusiveFromSlot, inclusiveToSlot) {
```
Thrillberg, line 75: This implementation is way better than my hacky `Imperial.passedThroughInvestor()`! 🤣

Thrillberg (commented):
This is only a partial review but this is looking very good to me!! Thanks again for your work on this!

app/javascript/lib/Entities/Rondel.js
GitGubM8: Hey, it worked without issues for the longest time and it never needed fixing

In my book, thats good code (until we do need to touch it)

app/javascript/lib/imperial.js
GitGubM8: I agree that we don't really need the action type, but the log would be helpful to players

Just to make sure I understand this correctly, if we're still logging this action through `annotatedLog`, isn't `annotatedLog` used to rebuild the game when the page is reloaded? Wouldn't we still need this to move to the next turn?

Thrillberg: Nope! The `gameLog` is used to build the game state and the `annotatedLog` is used to build the log at the bottom of the screen. So I want to keep `gameLog` as minimal as possible, giving the game logic only the bare minimum it needs to accurately calculate game state. The `annotatedLog` is a later addition which is really just for players convenience.

GitGubM8: Ahh I understand now, on it

app/javascript/lib/Entities/Board/Province.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+export const translateProvinceModel = (representation, allProvinces, allUnits, gameBoard) => {
+  // temporary until migration is complete
+  const oldProvinceModel = allProvinces.get(representation);
+  const geographicModel = gameBoard.graph.get(representation);
+
+  const province = new Province(
+    geographicModel.isOcean === false || geographicModel.isOcean === undefined,
```
Thrillberg: Super small point here but couldn't we do `new Province(!geographicModel.isOcean, representation)`?

app/javascript/lib/Entities/Board/Province.js
```diff
@@ -0,0 +1,65 @@
...
+  get isOccupiedByHostiles() {
+    return this.hostileUnits.size > 0;
+  }
+  get isNotOccupiedByHostiles() {
+    return this.isOccupiedByHostiles === false;
+  }
+}
+
+export const translateProvinceModel = (representation, allProvinces, allUnits, gameBoard) => {
+  // temporary until migration is complete
```
Thrillberg, line 40: Can you elaborate a little bit on how this will look when the migration is complete? Are we not going to have to pass all provinces, all units and the game board in each time?

app/javascript/lib/Entities/Board/Province.js
```diff
@@ -0,0 +1,65 @@
...
+  province.hasArmsFactory = oldProvinceModel.factory === 'armaments';
+  province.hasNavalFactory = oldProvinceModel.factory === 'shipyard';
+
+  for (const [occupyingNation] of allUnits) {
+    const { armies, fleets, friendly } = allUnits
+      .get(occupyingNation)
+      .get(representation);
+
+    if (armies > 0 || fleets > 0) {
+      const unitPool = friendly ? province.friendlyUnits : province.hostileUnits;
```
Thrillberg, line 56: Where is `unitPool` used?

app/javascript/lib/action.js
```diff
@@ -64,7 +66,7 @@ export default {
   initialize: makeAction("initialize", ["players", "soloMode", "variant", "baseGame"]),
   blockCanal: makeAction("blockCanal", []),
   bondPurchase: makeAction("bondPurchase", ["nation", "player", "cost", "tradeInValue"]),
-  buildFactory: makeAction("buildFactory", ["province"]),
+  buildFactory: makeAction("buildFactory", ["province", "player", "nationCosts", "playerCosts"]),
```
Thrillberg: I'm wondering if we can get away with omitting `nationCosts` and `playerCosts`, too. Here's my thinking:
- If the nation can afford the factory, the player MAY NOT supplement with player cash.
- If the nation cannot afford the factory, the player supplements with player cash.
- If both nation and player cannot afford the factory then this would be an invalid action.

The above would be mistaken if the player gets to _choose_ to supplement with their cash even though a nation can afford to build a factory. Is this ever possible? What do you think?

app/javascript/lib/imperial.js
```diff
@@ -1336,31 +1356,46 @@ export default class Imperial {
...
+            }),
+          );
+        } else {
+          this.annotatedLog.push(
+            Action.couldNotBuildFactory({
+              nation: this.currentNation,
+            }),
+          );
+
+          this.handlePassingThroughInvestor();
```
Thrillberg, line 1398: Is this redundant of line 194? (Although, 194, the `couldNotBuildFactory` action case should be removed anyway so this might be right, but just want to double check)

app/javascript/lib/imperial.js
```diff
@@ -1834,91 +1869,60 @@ export default class Imperial {
...
-      out.add(
-        Action.rondel({
-          nation,
-          cost: this.extraRondelCost(2),
-          slot:
-            rondelPositions[currentIndex + 5] || rondelPositions[distance + 5],
-        }),
-      );
-      out.add(
+  availableRondelActions(nationName) {
```
Thrillberg, line 1874: `nationName` is way better than `nation` here. Thanks!

app/javascript/lib/imperial.js
```diff
@@ -1834,91 +1869,60 @@ export default class Imperial {
...
+    for (const freeRondelSlot of nextAvailableFreeRondelSlots) {
+      availableRondelSlots.add(
         Action.rondel({
-          nation,
-          cost: this.extraRondelCost(3),
-          slot:
-            rondelPositions[currentIndex + 6] || rondelPositions[distance + 6],
+          nation: nationName,
+          cost: 0,
+          slot: freeRondelSlot.representation,
```
Thrillberg: Returning to the `.representation` vs `.toString` vs other naming question, this is a spot where I feel like `.toString()` would feel pretty natural. Or even doing `freeRondelSlot.name.toString()` would feel natural. Ok I'll drop it now 😄 `representation` is fine, I think I just need to get used to it! 😆

app/javascript/lib/imperial.js
```diff
@@ -1834,91 +1869,60 @@ export default class Imperial {
...
-    return out;
+
+    return availableRondelSlots;
   }
 
-  extraRondelCost(index) {
-    if (this.baseGame === 'imperial') {
-      return index * 2;
-    }
+  costPerPaidRondelAction(nation) {
```
Thrillberg: In our new Onion representation of the code, should this function live elsewhere? Maybe on the `AvailableSlots` use case?

app/javascript/lib/imperial.js
```diff
@@ -2332,7 +2336,8 @@ export default class Imperial {
     let bonus = this.playerBonusBeforeUnitMaintenanceCosts(nationName, taxes);
 
     if (this.baseGame === 'imperial2030' || this.baseGame === 'imperialAsia') {
-      const treasuryAmountAfterMaintenanceCosts = this.nations.get(nationName).treasury + taxes - this.unitMaintenanceCosts(nationName);
+      const treasuryAmountBeforeMaintenanceCosts = this.nations.get(nationName).treasury + taxes;
+      const treasuryAmountAfterMaintenanceCosts = treasuryAmountBeforeMaintenanceCosts - this.unitMaintenanceCosts(nationName);
```
Thrillberg, line 2331: This is a great clarification! 🙏

app/javascript/lib/imperial.test.js
```diff
@@ -3095,7 +3094,29 @@ describe('imperial', () => {
...
-              Action.buildFactory({ province: 'b' }),
+              Action.buildFactory({ province: 'b', player: 'player1', nationCosts: 5, playerCosts: 0 }),
+              Action.skipBuildFactory({ nation: Nation.AH, player: 'player1' }),
+              Action.undo({ player: 'player1' }),
+            ]),
+          );
+          expect(game.buildingFactory).toEqual(true);
+        });
+
+        test('player may fund nation to build factory if deficit', () => {
```
Thrillberg, line 3118: Thanks for adding a new test for this functionality!

app/javascript/src/components/Board.vue
```diff
@@ -141,7 +141,8 @@ export default {
       if (this.paused) return false;
 
       if (this.game.availableActions) {
-        return [...this.game.availableActions].every((action) => action.type === 'buildFactory' || action.type === 'skipBuildFactory');
+        return [...this.game.availableActions].every(
+          (action) => action.type === 'buildFactory' || action.type === 'skipBuildFactory');
```
Thrillberg: 👍

Thrillberg (commented):
Ok I think I've reviewed all the `.js` files! I'll submit this for now and turn to the `.vue` files next.

app/javascript/lib/imperial.js
GitGubM8: [d768fc4](https://github.com/Thrillberg/imperial/pull/711/commits/d768fc42d266139ac8332f5732885bf95536aeae)

app/javascript/lib/imperial.js
GitGubM8: Good catch, I also opted to reevaluate the costs again for similar reasons yet didnt do this part somehow

[166a5e0](https://github.com/Thrillberg/imperial/pull/711/commits/166a5e05f9dd4aa2fbd0d23cda0cee5f2a19c15f)

app/javascript/lib/imperial.js
GitGubM8: [166a5e0](https://github.com/Thrillberg/imperial/pull/711/commits/166a5e05f9dd4aa2fbd0d23cda0cee5f2a19c15f)

app/javascript/lib/Entities/Board/Province.js
GitGubM8: Yeah, personally I find the `!` kinda small and easy to miss, and therefore prefer `== false`, except that's not necessarily true here since `js` shenanigans

Will refactor to `! ... .isOcean` since it's clear what it's doing anyway

app/javascript/lib/Entities/Board/Province.js
GitGubM8: So far, my plan is to eventually retire both `game.provinces.get(provinceName)` and `board.provinces.get(provinceName)` with this model, where this entity would persist in cache on perhaps `Imperial.js`, and different usecases would change modify these `Provinces` directly

In other words, the `Provinces` will be constructed once in `Imperial.constructor`, and remain in place. When an action comes in and needs to refer to the `Province` entity, we only need to look up a dictionary once (similar to `Rondel.translate`), and will only interact with the entity from there on

That way, there would be little to no confusion in the logic that whenever we refer to a `province`, we are referring to the object, have all the associated data right away, instead of having to worry about whether it's actually the string id of the object or the real object itself

app/javascript/lib/Entities/Board/Province.js
GitGubM8: The line right below it I think, that way we no longer have to perform the same logic twice

The alternative is
```
if (friendly) {
  province.friendlyUnits.set(occupyingNation, armies + fleets);
} else {
  province.hostileUnits.set(occupyingNation, armies + fleets);
}
```
which feels a little redundant

app/javascript/lib/imperial.js
GitGubM8: I think it's a legitimate concern if the naming feels weird, going to try renaming it to `id` for now

It's mostly used to communicate with front-ends/packets, where they're going to refer to the object by ids instead of the pointer

app/javascript/lib/Entities/Board/Province.js
GitGubM8: [2a152d3](https://github.com/Thrillberg/imperial/pull/711/commits/2a152d342ba4c7edb3f493060703c859b8ceacd5)

app/javascript/lib/imperial.js
GitGubM8: [a6e9bfa](https://github.com/Thrillberg/imperial/pull/711/commits/a6e9bfa69e3e1cc1ad7604bf0fa247a5b967bbf0)
[98ff4a7](https://github.com/Thrillberg/imperial/pull/711/commits/98ff4a7d5dc19ab123d1268ed097915bf5c5b255)

app/javascript/lib/imperial.js
GitGubM8: Mostly to keep eslint from complaining about line length

app/javascript/lib/imperial.js
GitGubM8: Great point, I was going to leave that for later but since we're in the vicinity might as well do it now

app/javascript/lib/Entities/Board/Province.js
Thrillberg: Ah I didn't notice that setting on `unitPool` sets the property of `province`. Great!

app/javascript/lib/UseCases/Board/Provinces/Factory/Build.js
```diff
@@ -0,0 +1,12 @@
+export default class Build {
+  constructor(province) {
+    this.province = province;
+  }
+
+  buildArmsFactory() {
+    this.province.hasArmsFactory = true;
+  }
+  buildNavalFactory() {
+    this.province.hasNavalFactory = true;
```
Thrillberg, line 10: At some point we're going to run up against the discrepancy that I was calling these `shipyards` before. I kinda prefer the symmetry of `arms factory` and `naval factory` so it may be worth switching over. But just a heads up!

app/javascript/lib/imperial.js
GitGubM8: [33b0e98](https://github.com/Thrillberg/imperial/pull/711/commits/33b0e98d928c58369c8fa89469759623b392478a)

app/javascript/lib/UseCases/Board/Provinces/Factory/Build.js
GitGubM8: Yeah, I think for consistency we won't be switching the ids over, and will be keeping the term `armaments/shipyards`, but at least logic-wise from now on whenever we refer to a province, we can call them `Arms/NavalFactory`

GitGubM8: Also note that this usecase isn't actually used anywhere, again just in preparation for future refactors.

It looks super simple and unnecessary, but it provides a nice point of entry for future refactors where if we needed to add in additional logic regarding building a new factory, we'd know that this is the place to go to
Or perhaps later on we'll discover that this isn't really necessary at all and can retire it!

app/javascript/src/views/Game.vue
```diff
@@ -656,10 +658,7 @@ export default {
             action.payload.placements.forEach((placement) => {
               provinces.add(placement.province);
             });
-          } else if (
-            action.type === 'buildFactory'
-            && this.game.buildingFactory
-          ) {
+          } else if (action.type === 'buildFactory' && this.game.buildingFactory) {
```
Thrillberg, line 659: Maybe we should omit this file from the diff since I think it's just linting changes/complaints from GitHub.

app/javascript/src/getGameLog.js
```diff
@@ -3,58 +3,34 @@ import { Nation, Nation2030, NationAsia } from '../lib/constants';
...
   // The following map only exists because of our custom Nation type, which
   // has weirdness when we attempt nation.when() in the setup file.
+  let nations;
+  if (baseGame === 'imperial') {
+    nations = Nation;
+  } else if (baseGame === 'imperial2030') {
+    nations = Nation2030;
+  } else if (baseGame === 'imperialAsia') {
+    nations = NationAsia;
+  }
```
Thrillberg, line 13: Lol this is way cleaner. Thank you!

app/javascript/src/components/Board.vue
```diff
@@ -141,7 +141,9 @@ export default {
       if (this.paused) return false;
 
       if (this.game.availableActions) {
-        return [...this.game.availableActions].every((action) => action.type === 'buildFactory' || action.type === 'skipBuildFactory');
+        return [...this.game.availableActions].every(
+          (action) => action.type === 'buildFactory' || action.type === 'skipBuildFactory',
+        );
```
Thrillberg, line 146: Actually, can we remove this from the diff as well?

Thrillberg (commented):
Vue files look good over all (a lot of linting in there, I guess, but that's ok). One thing I wonder about is how the user interaction goes when they are about to finance a factory from their own cash. It looks like it'll be more or less invisible as is, which is ok for now but we may want to message more clearly when the game will take money from the players cash vs all $5m from the nation's treasury. What do you think?

app/javascript/lib/imperial.js
GitGubM8: [a246c6e](https://github.com/Thrillberg/imperial/pull/711/commits/a246c6ecbc4ae05ff4504869f73b34720fcb045c)

app/javascript/lib/imperial.js
GitGubM8: I thought the `Tick` is called after a user input, so this was necessary since there's no further user input required

GitGubM8:
> Vue files look good over all (a lot of linting in there, I guess, but that's ok). One thing I wonder about is how the user interaction goes when they are about to finance a factory from their own cash. It looks like it'll be more or less invisible as is, which is ok for now but we may want to message more clearly when the game will take money from the players cash vs all $5m from the nation's treasury. What do you think?

Yeah I was thinking about either adding a static message, or more complicatedly but clearly, calculating their costs, when they hover over Rondel

app/javascript/src/components/Board.vue
GitGubM8: I think it needed to be on an extra line cause lint was complaining, lint likes it the way it is now 

Tangentially related
[0cca894](https://github.com/Thrillberg/imperial/pull/711/commits/0cca8940a4eca8b10932615065951e453a9db159)

app/javascript/src/views/Game.vue
GitGubM8: Eh, this line was under the limit unless you preferred it the way it was before

Up to you, I thought it looked neater when on a single line
[0cca894](https://github.com/Thrillberg/imperial/pull/711/commits/0cca8940a4eca8b10932615065951e453a9db159)

app/javascript/lib/action.js
GitGubM8: Yeah we dont really need them other than for display purposes, and this was the quickest way to pass the data along so that the frontend didn't have to touch any game logic

Up to you, my only intention was to avoid frontend having to pick up the usecases and calculate the costs. When it comes back, the costs are re-calculated to avoid any malicious stuff anyway

GitGubM8:
Added a simple hint for now

image
[465ed85](https://github.com/Thrillberg/imperial/pull/711/commits/465ed854adb7973a8d366105087c47a308ced029)

GitGubM8:
The best way is probably to add a new green button that gives them the option "Fund nation" / "Skip Building" before they pick where to build the factory

But at this point I'd like to wrap this up for now and wait for anyone to complain about it if any

Thrillberg (approved):
Looks good to me! Thank you, again!!

PR #719, opened 2023-03-25
Linting. Nothing to see here.

Thrillberg:
@GitGubM8 Theoretically this PR should only have linting updates to files that are touched in the factory finance branch (except for `Game.vue`, which I still think we can leave untouched for now). Does this make it easier to graft your PR onto and end up with a lower count of changed lines? It might be a drop in the bucket but I figured at least for review purposes it's easier to have linting separate.

### DIFF
--- a/app/javascript/lib/action.js
+++ b/app/javascript/lib/action.js
@@ -1,4 +1,4 @@
-import { memoize } from "./memo.js";
+import { memoize } from './memo';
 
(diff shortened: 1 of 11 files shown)


GitGubM8 (approved):

GitGubM8:
Yeah let's merge this so that I can do some nightly testing and revert if necessary

PR #721, opened 2023-03-25
Factory Finance

GitGubM8:


GitGubM8:
@Thrillberg 
Ironically, the source of the failures last night were [bf5a777](https://github.com/Thrillberg/imperial/pull/721/commits/bf5a77753ab2cd92541c12d9f9fcd5e1a8c11110)

For the record, I did say the original code was good code since it never failed :)

PR #722, opened 2023-03-25
New YarnLock

GitGubM8:
Had some issues with dependencies this morning, worked after I ran `yarn`

Here's the `yarn.lock` file if we want it, I have no idea if the dependencies it generated are actually correct


Thrillberg:
Admittedly, I haven't looked into `yarn` vs `npm` very much but my impression was that we'd want to choose one or the other and not both. So far this codebase has not been using `yarn` so I think it's best to keep it that way. What were the dependency issues? It probably arose because I added a types package last night as part of the linting work.

GitGubM8:
I see, I don't understand it too much either, but I read that yarn is built in top of npm and helps manage dependencies for the entire repo

Feel free to close this PR, you're more familiar with this stuff anyway, and I got the dependencies resolved on my side already

Thrillberg:
Yeah let's close it for now. We can look more deeply into it at another time if we need to.

PR #725, opened 2023-03-25
Move logging to Logtail

Thrillberg:
There's some linting junk in here too, but not too much. What do you think of this @GitGubM8 ? Is this worth doing? If you want access to Logtail I can give that, but totally understood if you don't!

This screenshot shows the kind of log it sends to Logtail. Any other fields we should collect? (The screenshot doesn't show the error correctly because I didn't actually submit an invalid move locally to produce this. But I _think_ it'll work in prod!)

Screenshot 2023-03-25 at 2 05 14 PM

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,8 +28,9 @@
             "object": true
         }],
         "vue/multi-word-component-names": ["error", {
-            "ignores": ["Board", "Game", "Province", "Flag", "Bond", "Player", "Rondel", "About"]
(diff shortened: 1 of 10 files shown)


GitGubM8:
Looks great, you'll have to show me how to use it in a bit

GitGubM8 (approved):

Thrillberg:
Sweet! I'm also restricting it so it only logs in prod. We'll get less noise that way. Then I'll merge this.

PR #726, opened 2023-03-26
Import/Export games

Thrillberg:
So far all this does is allow someone to download a game by hitting a certain endpoint (`/exports/[:game_id]`). Things left to do:

1. Build a FE for submitting a game log JSON (file upload? text paste?). Be sure that only certain people have access to this and that it is disabled on prod. Should only be needed on development environment.
2. Build a way to reconstruct a game from just a log (???) Do we need more than just the log? What will player names be? Does it matter?
3. ???
4. Profit! (but don't cuz we didn't design this board game in the first place)

UPDATE: Everything should be working but we just need to test it all!

### DIFF
--- a/app/javascript/lib/imperial.js
+++ b/app/javascript/lib/imperial.js
@@ -1,4 +1,3 @@
-import { Logtail } from '@logtail/browser';
 import { translateProvinceModel } from './Entities/Board/Province';
 import Imperial2030Game from './Entities/Imperial2030Game';
 import ImperialAsiaGame from './Entities/ImperialAsiaGame';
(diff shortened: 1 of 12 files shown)


app/javascript/lib/imperial.js
```diff
@@ -38,7 +37,6 @@ export default class Imperial {
   }
 
   constructor(board) {
-    this.logtail = new Logtail('3bdHcA8P3mcww2ojgC5G8YiT');
```
GitGubM8, line 41: Is this intentional?

GitGubM8 (approved):
Lets see if this works!

Doesn't look like it'd break anything even if it doesnt so thats good

app/javascript/lib/imperial.js
Thrillberg: Yep, I had accidentally left it in here from before. I never use it in this file so it's not necessary.

PR #729, opened 2023-03-27
Refactored out Logger Service

GitGubM8:


GitGubM8:
Retired the legacy tests cause cant be bothered to update them

app/javascript/lib/auction.test.js
```diff
@@ -1,9 +1,22 @@
...
+
 import Auction from './auction';
 import Imperial from './imperial';
 import { Bond, Nation } from './constants';
 import auctionSetup from './auctionSetup';
 
+import BoardEurope from './board';
+
+const ImperialFromLog = (log) => {
+  const game = new Imperial(BoardEurope, new Logger('unittests', 'stub-test-id'));
```
Thrillberg: Does `Imperial` need to require a `logger`? What if we pass nothing? We really only want the logger in production anyway.

app/javascript/lib/imperial.test.js
```diff
@@ -1,8 +1,14 @@
 import Action from './action';
+
+import Logger from '../src/Logger';
+
 import { Bond, Nation } from './constants';
 import GameBoard from './gameBoard';
+
 import Imperial from './imperial';
 
+const newImperialGame = (board) => new Imperial(board, new Logger('unittests', 'stub-test-id'));
```
Thrillberg: Same q, can we remove the Logger?
Hmm and without the logger I'm not sure what benefit this refactor gets us.

app/javascript/lib/imperial2030.test.js
```diff
@@ -1,8 +1,13 @@
 import Action from './action';
+
+import Logger from '../src/Logger';
+
 import GameBoard from './gameBoard';
 import Imperial from './imperial';
 import { Nation2030 } from './constants';
 
+const newImperialGame = (board) => new Imperial(board, new Logger('unittests', 'stub-test-id'));
```
Thrillberg: Here too.

app/javascript/lib/imperialWithoutInvestorCard.test.js
```diff
@@ -1,7 +1,20 @@
...
+
+import Logger from '../src/Logger';
+
 import Imperial from './imperial';
+
+import BoardEurope from './board';
 import { Nation } from './constants';
 
+const ImperialFromLog = (log) => {
+  const game = new Imperial(BoardEurope, new Logger('unittests', 'stub-test-id'));
```
Thrillberg: And here.

app/javascript/lib/schnelleinsteig_legacy_tests.js
```diff
@@ -1,1847 +0,0 @@
-// This file is not run anymore because it is out of date and not worth
```
Thrillberg: Sad to see this go, but it's ok! 😢 Thanks for removing it!

app/javascript/src/Logger.js
```diff
@@ -0,0 +1,34 @@
...
+    this.#environment = environment;
+    this.#gameId = gameId;
+
+    this.#logtail = new Logtail('3bdHcA8P3mcww2ojgC5G8YiT');
+  }
+
+  error(message, data) {
+    data.gameId = this.#gameId;
+
+    switch (this.#environment) {
```
Thrillberg: I wonder if we can use a word other than "environment". I think of "environment" as just being "development", "test", and "production". But if we have a "replay" one then that seems like a different paradigm. It's still in `production`. What if we call it a "scope"? I'd want to use "mode" but we already have game mode elsewhere, I think. What do you think?

app/javascript/src/Logger.js
```diff
@@ -0,0 +1,34 @@
...
+      case 'production':
+        this.#logtail.error(message, data);
+        break;
+
+      case 'replay':
```
Thrillberg, line 24: Should we populate this with `this.#logtail.error(message, data);` like we do in `production`?

app/javascript/src/Logger.js
```diff
@@ -0,0 +1,34 @@
...
+      case 'unittests':
```
Thrillberg: I don't imagine we'll want to use the logger in the test environment (as per my comments above, too).

app/javascript/src/views/Game.vue
```diff
@@ -783,9 +775,15 @@ export default {
       }
 
       const { log } = this.game;
-      const { board } = this.game;
+      const { baseGame } = this.game;
 
-      this.game = Imperial.fromLog(log, board);
+      const game = new Imperial(this.board, new Logger('replay', this.gameData.id));
```
Thrillberg, line 780: Love this use of `replay`!

app/javascript/lib/imperial.js
```diff
@@ -32,16 +30,12 @@ import AvailableSlots from './UseCases/Rondels/SlotSelection/AvailableSlots';
...
-    log.forEach((entry) => game.tick(entry));
-    return game;
-  }
+  #logger;
+  #game;
 
-  constructor(board) {
-    this.logtail = new Logtail('3bdHcA8P3mcww2ojgC5G8YiT');
+  constructor(board, logger) {
+    this.#logger = logger;
```
Thrillberg: What happens if we don't pass in a logger?

Thrillberg (commented):
Looks good to me! Mostly just a question of making the logger optional. I think some of the test refactors aren't really necessary/helpful but happy to hear the rationale!

app/javascript/lib/auction.test.js
GitGubM8: No, you're right, it doesn't need a logger

I guess I wanted to avoid having to null-check every time we call `this.#logger` inside `Imperial.js`, and having a stub logger was the best solution for that.

Writing wrappers methods within `Imperial.js` to perform the null-check seems a bit redundant, since that'd imply we'd need a new wrapper method for every method defined in `Logger`?

app/javascript/src/Logger.js
GitGubM8: Yeah I wasn't too keen on having `replay` or even `unittest` as an environment, and didn't have time to think up of a better design

Let's go with the rename of `scope`, I think it quite accurately describes the purpose of this without having to redesign the whole approach. Good naming

app/javascript/lib/imperial.js
GitGubM8: Again, just null-checks required before each use of logger

app/javascript/lib/schnelleinsteig_legacy_tests.js
GitGubM8: I'd have liked to keep it around as well :c
keeping it compilable was a pain tho

Unless I was actually wrong and it didn't actually need to be compilable in the first place

app/javascript/lib/auction.test.js
GitGubM8: I see in javascript there's actually optional chaining `?.` too, that simplifies things

Lets clear this out then

app/javascript/lib/imperial.js
Thrillberg: How about if we change this to:

```javascript
this.#logger = logger || new Logger();
```

and this should fall through to the `default` case in the env check on `Logger.js`
```javascript
switch (this.#environment) {
  ...
  default:
    console.error(message, data);
    break;
}
```
Then we can remove this from the tests. Would that work?

app/javascript/lib/schnelleinsteig_legacy_tests.js
Thrillberg: yeah I don't think these get run in CI or anything, do they? but they are def gonna fail at this point.

app/javascript/lib/auction.test.js
Thrillberg: Yeah that could work, too. Check my other comment about making a stub Logger as the default, which would not actually send logs through Logtail. If that works, it might be preferable.

app/javascript/lib/imperial.js
GitGubM8: That could work too, do we want it to log in the tests tho, since it already logs on CI?

Or maybe the default case should be a silent fail, altho Im not as keen on silent failures.

Otherwise I'll continue with my plan of just `this?.#logger` everything

Thrillberg: >do we want it to log in the tests tho, since it already logs on CI?

Not sure I understand. This would log locally in tests, which is probably a good thing, right? Like, if a test is passing but encounters an error anyway, it would be good to see that in the terminal output. Not sure what you mean by it already logs on CI.

>Or maybe the default case should be a silent fail, altho Im not as keen on silent failures

I feel the same way but I'm ok with it in this case if that seems better than `console.error`ing as above. Theoretically we shouldn't be seeing errors on tests that are passing. But you never know...!

GitGubM8: I see, I've never ran the tests locally, but yeah in that case it'd be more beneficial than detrimental, lets go with this plan than

app/javascript/src/Logger.js
GitGubM8: Im not sure, wouldn't that mean that we get a duplicate `logtail.error` everytime someone rewinds the game?

Thrillberg: I think not because the instance of `Logger` after they've clicked the "back" button (or similar) would have the `replay` scope and not the `production` scope, right? I think the code as written here will swallow the error if they are rewinding the game.

app/javascript/lib/imperial.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

app/javascript/src/Logger.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

app/javascript/src/Logger.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

app/javascript/lib/imperialWithoutInvestorCard.test.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

app/javascript/lib/imperial2030.test.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

app/javascript/lib/imperial.test.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

GitGubM8: Oh, there's also a tiny refactor where I removed `static Imperial.fromLogs` and made it a class method `new Imperial().tickFromLog()`

which helps a tiny bit with making sure that users of `tickFromLog` will have to go through the actual constructor and provide its required dependencies

IE. future changes to parameters `Imperial.constructor` will no longer propagate through `Imperial.fromLog`

app/javascript/lib/auction.test.js
GitGubM8: [76f61f0](https://github.com/Thrillberg/imperial/pull/729/commits/76f61f00f58e876e15be9e5e2d7e0ddfdbd8a432)

app/javascript/src/Logger.js
GitGubM8: So no changes here then?

app/javascript/lib/imperial.test.js
Thrillberg: Yeah I think that refactor is good but changing `new Imperial(board)` to `newImperialGame(board)` doesn't make a lot of sense to me, if `newImperialGame()` isn't bringing any other logic in or anything. (This is currently the case in `imperial.test.js`)

app/javascript/src/Logger.js
Thrillberg: Is the goal to suppress logs from replay? If so, then this is fine as-is. But if we want logs from replay then we want it to act like production above, right? I'm ok either way, though. 😄 

EDIT: Ok I think I had misunderstood the intent. yes this would cause duplicate logs. Let's keep as-is!

app/javascript/lib/auction.test.js
Thrillberg: There's still a logger here. 😄

app/javascript/lib/imperial.test.js
GitGubM8: Oh no wait, I was mistaken, thought you're referring to `newImperialFromLog`, yeah, in that case now that we've removed the logger this is unnecessary

Reverting

app/javascript/lib/auction.test.js
GitGubM8: Good catch

app/javascript/lib/imperial.test.js
GitGubM8: [7cdd568](https://github.com/Thrillberg/imperial/pull/729/commits/7cdd568bd4c56ec463bf0db35fd5482eb7dfe399)

app/javascript/lib/auction.test.js
GitGubM8: [a152579](https://github.com/Thrillberg/imperial/pull/729/commits/a152579aea30692747b82fa28d96d29a6950a526)
[07df3a1](https://github.com/Thrillberg/imperial/pull/729/commits/07df3a17aedc0229a4fbcd0ef4545be64416a437)

Thrillberg (approved):
Awesome, looks great! If you want to delete the legacy tests, that's ok, although if it's possible to retain them that would be nice. Early on I was so pleased with myself when I discovered that a good way to develop the game was just to follow the sample game of 4 turns and make that playable. So I made a whole test suite out of it and just let the pamphlet be my guide!

Anyway, I think they don't need to compile or run so it's purely for sentimental reasons that I want to keep it around!

GitGubM8:
> 

Ah, in that case let's keep it them, I'd rather have them around for easier reference too

PR #731, opened 2023-03-28
New nation initialization

GitGubM8:
## Comments:
- Tested and verified particularly the undo Action. Apparently that's also what causes the next action to be duplicated or something, causing all the invalid logs on production, which it rightfully ignored.

### DIFF
--- a/app/javascript/lib/Entities/AbstractImperialGame.js
+++ b/app/javascript/lib/Entities/AbstractImperialGame.js
@@ -8,14 +8,42 @@ export default class AbstractImperialGame extends Entity {
   }
 
   #rondel;
+  #availableFreeRondelSlotCount = 3;
+  #availablePaidRondelSlotCount = 3;
 
-  constructor(id) {
+  #nations;
+  #nationOrder;
+
+  constructor(id, nations, nationOrder) {
     super(id);
 
     this.#rondel = new Rondel();
+
+    this.#nations = nations;
+    this.#nationOrder = nationOrder;
   }
 
   get rondel() {
     return this.#rondel;
+  }
+
+  get availableFreeRondelSlotCount() {
+    return this.#availableFreeRondelSlotCount;
+  }
+  get availablePaidRondelSlotCount() {
+    return this.#availablePaidRondelSlotCount;
+  }
+
+  nationIdToEntity(nationId) {
+    return this.#nations.get(nationId);
+  }
+
+  nationTurnAfter(nation) {
+    let nationIndex = this.#nationOrder.indexOf(nation);
+
+    nationIndex += 1;
+    nationIndex %= this.#nationOrder.length;
+
+    return this.#nationOrder[nationIndex];
   }
 }
--- a/app/javascript/lib/Entities/Imperial2030Game.js
+++ b/app/javascript/lib/Entities/Imperial2030Game.js
@@ -1,7 +1,67 @@
 import AbstractImperialGame from './AbstractImperialGame';
 
+import Nation from './Nations/Nation';
+
 export default class Imperial2030Game extends AbstractImperialGame {
   static get classId() {
     return 'imperial2030';
+  }
+
+  static get RussiaId() {
+    return 'RU';
+  }
+  static get ChinaId() {
+    return 'CN';
+  }
+  static get IndiaId() {
+    return 'IN';
+  }
+  static get BrazilId() {
+    return 'BR';
+  }
+  static get UnitedStatesId() {
+    return 'US';
+  }
+  static get EuropeanUnionId() {
+    return 'EU';
+  }
+
+  constructor(id) {
+    const nations = new Map();
+    nations.set(Imperial2030Game.RussiaId, new Nation(Imperial2030Game.RussiaId));
+    nations.set(Imperial2030Game.ChinaId, new Nation(Imperial2030Game.ChinaId));
+    nations.set(Imperial2030Game.IndiaId, new Nation(Imperial2030Game.IndiaId));
+    nations.set(Imperial2030Game.BrazilId, new Nation(Imperial2030Game.BrazilId));
+    nations.set(Imperial2030Game.UnitedStatesId, new Nation(Imperial2030Game.UnitedStatesId));
+    nations.set(Imperial2030Game.EuropeanUnionId, new Nation(Imperial2030Game.EuropeanUnionId));
+
+    const nationOrder = new Array(6);
+    nationOrder[0] = nations.get(Imperial2030Game.RussiaId);
+    nationOrder[1] = nations.get(Imperial2030Game.ChinaId);
+    nationOrder[2] = nations.get(Imperial2030Game.IndiaId);
+    nationOrder[3] = nations.get(Imperial2030Game.BrazilId);
+    nationOrder[4] = nations.get(Imperial2030Game.UnitedStatesId);
+    nationOrder[5] = nations.get(Imperial2030Game.EuropeanUnionId);
+
+    super(id, nations, nationOrder);
+  }
+
+  get Russia() {
+    return this.nationIdToEntity(Imperial2030Game.RussiaId);
+  }
+  get China() {
+    return this.nationIdToEntity(Imperial2030Game.ChinaId);
+  }
+  get India() {
+    return this.nationIdToEntity(Imperial2030Game.IndiaId);
+  }
+  get Brazil() {
+    return this.nationIdToEntity(Imperial2030Game.BrazilId);
+  }
+  get UnitedStates() {
+    return this.nationIdToEntity(Imperial2030Game.UnitedStatesId);
+  }
+  get EuropeanUnion() {
+    return this.nationIdToEntity(Imperial2030Game.EuropeanUnionId);
   }
 }
--- a/app/javascript/lib/Entities/Nations/Nation.js
+++ b/app/javascript/lib/Entities/Nations/Nation.js
@@ -1,0 +1,10 @@
+import Entity from '../Entity';
+
+export default class Nation extends Entity {
+  constructor(id) {
+    super(id);
+
+    this.residingRondelSlot = null;
+    this.powerPoints = 0;
+  }
+}
--- a/app/javascript/lib/UseCases/Rondels/MoveToSlot.js
+++ b/app/javascript/lib/UseCases/Rondels/MoveToSlot.js
@@ -29,7 +29,7 @@ export default class MoveToSlot {
     return this.#costs;
   }
 
-  forceMoveNation(nation, toRondelSlot) {
+  static forceMoveNation(nation, toRondelSlot) {
     if (nation.residingRondelSlot) {
       nation.residingRondelSlot.residingNations.delete(nation);
     }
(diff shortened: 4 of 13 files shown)


app/javascript/lib/Entities/AbstractImperialGame.js
```diff
@@ -8,14 +8,42 @@ export default class AbstractImperialGame extends Entity {
...
+  }
+  get availablePaidRondelSlotCount() {
+    return this.#availablePaidRondelSlotCount;
+  }
+
+  nationIdToEntity(nationId) {
+    return this.#nations.get(nationId);
+  }
+
+  nationTurnAfter(nation) {
```
GitGubM8, line 41: Next minor refactor I'll probably start moving out `nextNation` logic from `imperial.js`
After that will probably be the `Taxation` rondel slot

We'll probably start seeing decent reductions to `imperial.js` once those happen

app/javascript/lib/UseCases/Rondels/MoveToSlot.js
```diff
@@ -0,0 +1,62 @@
...
+  constructor(fromRondelSlot, toRondelSlot) {
+    super(`Attepting to make invalid move from ${fromRondelSlot?.id} to ${toRondelSlot.id}.`);
+
+    this.name = 'InvalidMoveError';
+  }
+};
+
+export default class MoveToSlot {
+  static get InvalidMoveError() {
+    return InvalidMoveError;
```
GitGubM8, line 14: Im inclined to disagree and ignore lint's recommendation of `Only 1 class per file`

With this setup, we can write nested classes, which indicates a high degree of relationship between the 2 classes. As an external user, its access would look like
```
import MoveToSlot from './MoveToSlot'

try {
...
} catch (error) {
  if (error instanceof MoveToSlot.InvalidMoveError) {
    ...
  }
}
```
as opposed to
```
import InvalidMoveError from './InvalidMoveError'

try {
...
} catch (error) {
  if (error instanceof InvalidMoveError) {
    ...
  }
}
```
which doesn't really hint to me its relationship with others modules unless the reader is already very familiar with the codebase

Since extensions of error classes tend to be very light and seldomly reused, creating a new file for every one of them feels like clutter on the file hierarchy level as well

Would like to also hear your thoughts before progressing

app/javascript/lib/UseCases/Rondels/MoveToSlot.js
```diff
@@ -0,0 +1,62 @@
...
+
+  get availableSlots() {
+    return this.#availableSlots;
+  }
+  get slotDistanceCosts() {
+    return this.#costs;
+  }
+
+  forceMoveNation(nation, toRondelSlot) {
+    if (nation.residingRondelSlot) {
```
GitGubM8, line 33: While we're on the subject of lint, Im not too keen on it forcing me to declare static methods either, but Im half willing to let it slide.

Usecases tend to be more volatile, and can easily be changed to need class variables on a dime (or vice-versa). Prematurely declaring it static causes more annoyances down the road should that happen imo, but it's not too bad to fix either since it's usually just search and replace

GitGubM8: [a07bfc6](https://github.com/Thrillberg/imperial/pull/731/commits/a07bfc69ce5c855a6936c4d64057169698395e72)

app/javascript/lib/UseCases/Rondels/MoveToSlot.js
GitGubM8: Would like to hear your thoughts before doing anything

app/javascript/lib/Entities/AbstractImperialGame.js
```diff
@@ -8,14 +8,42 @@ export default class AbstractImperialGame extends Entity {
...
 
   #rondel;
+  #availableFreeRondelSlotCount = 3;
+  #availablePaidRondelSlotCount = 3;
 
-  constructor(id) {
+  #nations;
+  #nationOrder;
+
+  constructor(id, nations, nationOrder) {
```
Thrillberg, line 17: `id` here is the game type, like `Imperial2030Game`, right? If so can we call it something other than `id`? maybe `baseGameId`?

app/javascript/lib/Entities/AbstractImperialGame.js
```diff
@@ -8,14 +8,42 @@ export default class AbstractImperialGame extends Entity {
...
+
+  nationIdToEntity(nationId) {
+    return this.#nations.get(nationId);
+  }
+
+  nationTurnAfter(nation) {
+    let nationIndex = this.#nationOrder.indexOf(nation);
+
+    nationIndex += 1;
+    nationIndex %= this.#nationOrder.length;
```
Thrillberg, line 45: What is this line doing?

app/javascript/lib/Entities/Imperial2030Game.js
```diff
@@ -1,7 +1,67 @@
...
+    nationOrder[2] = nations.get(Imperial2030Game.IndiaId);
+    nationOrder[3] = nations.get(Imperial2030Game.BrazilId);
+    nationOrder[4] = nations.get(Imperial2030Game.UnitedStatesId);
+    nationOrder[5] = nations.get(Imperial2030Game.EuropeanUnionId);
+
+    super(id, nations, nationOrder);
+  }
+
+  get Russia() {
+    return this.nationIdToEntity(Imperial2030Game.RussiaId);
```
Thrillberg, line 50: Ah I see, so it's only storing strings in the `nationOrder`, for example, and then converting to the entity at the last minute on access. Makes sense!

app/javascript/lib/Entities/Nations/Nation.js
```diff
@@ -0,0 +1,10 @@
+import Entity from '../Entity';
+
+export default class Nation extends Entity {
+  constructor(id) {
```
Thrillberg, line 4: I'm curious if it makes sense to have this `Nation` entity define `idToEntity` instead of the Game entities having `nationIdToEntity` functions. I don't think it makes much difference but just wanted to check if you'd considered that.

app/javascript/lib/UseCases/Rondels/MoveToSlot.js
Thrillberg: Yep, I'll merge a pr that ignores this linting error. In general I have no particular allegiance to the linter rules as they are so if you want to override them, feel free.

Thrillberg (commented):
This is a partial review with a couple comments/questions! I'll try to review the rest later today. Thanks!!

app/javascript/lib/Entities/AbstractImperialGame.js
GitGubM8: Not sure if js supports out of bound indices, just wrapping it around to make sure it's between `0 to length -1`

app/javascript/lib/Entities/AbstractImperialGame.js
GitGubM8: Originally I considered each game's id `abcde-4654-hijkl` to be part of the domain logic, and was leaving this id to populate that, but in your other comment you considered it as more of a web-concern, so Im not too sure how to utilize this 'id'

For now there hasn't been any need for it yet so it's just hanging around

I suppose my argument it for being a domain concern is that eventually users will need to refer to a specific game (ie. loading a past game, making moves across different games), and that's where it is referred to.
On the other hand I do agree that I dont ever envision it being used inside game logic so, is it really needed?

Still thinking about it, probably leaning towards your suggestion

app/javascript/lib/Entities/Nations/Nation.js
GitGubM8: If we do that that would require a static registry for Nations to look up and reference upon conversion. That would also mean that we'd need to populate this static registry in an initialization method somewhere, chances are, also in `xxxImperialGame`

I think the placement of the registry in `xxxImperialGame` is correct since not only is it the creator/destructor of all these nations, which naturally makes it liable for the management of `Nations` entities, but also the fact that there are duplications in nation names across games (IE. Russia appears in all 3 games iirc), which could make it confusing to access.

For example, if Im querying `Nation.idToEntity('ru')`, will I get `ImperialEurope.Russia` back, or will I get `Imperial2030.Russia` back

GitGubM8: One could argue that as long as the scope registers/unregisters the nations properly, then it won't be a problem either, since only 1 scope is active at a time

At which point, again, isn't `xxxImperialGame` effectively that scope? Not to mention that we'd be subtlely placing a limitation of only allowing us to only be able to process 1 game at a time (not that we currently process more than 1, but that could be possible in the future)

app/javascript/lib/Entities/AbstractImperialGame.js
Thrillberg: Yeah I'm thinking it would be nice to stick with a strict division between the web stuff and the game logic stuff. So, "users" for example aren't a concern in the game logic, only "players" are. Likewise, the game's id is a web concern. A single instance of `Imperial`, the game, doesn't think of itself as having an id. I'd be willing to bend on this for a good argument but I'm not currently seeing it.

One shorthand I have for it all is files that end in `.js` are game logic and not concerned with the web and files that end in `.vue` are web. (ofc, I break this division all the time by putting game logic in the Vue files but that is another story. Tbh I think the Vue files are arguably in more need of refactoring than the JS files)

app/javascript/lib/Entities/Nations/Nation.js
Thrillberg: Great, sounds good to me! Let's keep as-is!

app/javascript/lib/imperial.js
```diff
@@ -128,6 +129,12 @@ export default class Imperial {
         if (this.auction?.inAuction) {
           Object.assign(this.auction, this.oldAuctionState);
         }
+
+        const currentNation = this.nations.get(this.currentNation);
+        if (currentNation.rondelPosition) {
+          const lastRondelSlot = this.#game.rondel.idToEntity(currentNation.rondelPosition);
```
Thrillberg: I know `this.#game` was already in this file but I must have missed it before. Previously, my understanding of the instantiation of an `Imperial` was that it was a game itself. Now it looks like a game is one of its properties. Is that right? If that's the case, what is an `Imperial`? (ie, this class, the container of the game) Maybe we should call it `ImperialGameRunner` or something?

Thrillberg (approved):
Only one comment here! And it's really more of a question. Looks good to me! Thanks for the refactor.

Thrillberg:
I've approved, but we need a clean test suite to pass before merging of course!

GitGubM8:
@Thrillberg An example of the scenario I mentioned in https://github.com/Thrillberg/imperial/pull/733 is with the CI's right now, where it can get pretty rough to sift through all the logs and identify exactly which unittests are failing

Thrillberg:
True but if you scroll to the bottom you'll see that 34 tests are failing so I think these are all legit and should be fixed. Ofc, it does _add_ to the overall length that the errors are there in addition to the regular test failures. But once tests are passing it should be short again.

app/javascript/lib/imperial.js
GitGubM8: Yeah, Im actually treating `imperial.js` as more of a coordinator class (and for now have named as such in the unittests)

Haven't gotten around to actually renaming the file yet since there's a lot of other classes that import `imperial.js`, and doing so would incur massive diffs

Next PR?

Thrillberg: Sure sounds good to me! There's also this issue which I stumbled upon earlier today in one of my games: https://github.com/Thrillberg/imperial/issues/732 I can work on it but if you'd rather, that's fine too.

GitGubM8: Naw, going to wrap all this up first before opening up another front

app/javascript/lib/imperial.test.js
```diff
@@ -2042,680 +2056,705 @@ describe('imperial', () => {
           return game;
         };
 
-        ['maneuver1', 'production1', 'factory', 'taxation'].forEach(
```
GitGubM8, line 2045: Unfortunately I had to reduce this to only `maneuver1` case

The new `MoveToRondelSlot` usecase actually enforces taking money from players when moving extra distances, and it'd be tricky to rewrite these to account for that

Thrillberg (approved):
Thanks for your work on this refactor! This PR has gotten a little unwieldy and I can't carefully review all of it. So I'll approve for now and we can revert if things go obviously wrong. In the future, can we try to reduce PR sizes to be far below 1k lines? I'm more accustomed/comfortable with PRs of up to a couple hundred lines change. Thanks!

GitGubM8:
Yeah, it's meant to be a lighter refactor already :T, I think a lot of it was actually just indentation changes in unittest is all

PR #733, opened 2023-03-28
Fix some noisy tests

Thrillberg:
This PR does a couple things that should help:
- Introduce a `'silent'` Logger so that when we test purposely-invalid moves it doesn't clutter up logs.
- Fix some tests that were silently failing. (For example, some `endManeuver`s are removed because when they were first put in we didn't automatically end a player's maneuver turn when they ran out of valid maneuvers. Once that rule came online, these tests were not updated because they were silently dropping those invalid actions)

### DIFF
--- a/app/javascript/lib/imperial.test.js
+++ b/app/javascript/lib/imperial.test.js
@@ -3,6 +3,8 @@ import Action from './action';
 import { Bond, Nation } from './constants';
 import GameBoard from './gameBoard';
 
+import Logger from '../src/Logger';
+
 import Imperial from './imperial';
 
 const cloneUnits = (units) => {
@@ -633,7 +635,6 @@ describe('imperial', () => {
           game.tick(
             Action.rondel({ slot: 'maneuver2', cost: 0, nation: Nation.AH }),
           );
-          game.tick(Action.endManeuver());
           game.tick(
             Action.rondel({ slot: 'taxation', cost: 0, nation: Nation.IT }),
           );
@@ -2222,7 +2223,7 @@ describe('imperial', () => {
                 // Set AH's rondel position to be something *before* investor
                 game.nations.get(Nation.AH).rondelPosition = startingPosition;
 
-                // The investor slot lies between 'maneuver1' and 'maneuver2'
+                // The investor slot lies between 'maneuver1' and 'production2'
                 game.tick(
                   Action.rondel({
                     slot: 'production2',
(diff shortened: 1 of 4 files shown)


Thrillberg:
@GitGubM8 Shall we merge this before https://github.com/Thrillberg/imperial/pull/731? That way we can clean up the npm test output on CI.

app/javascript/lib/imperial.test.js
```diff
@@ -4624,7 +4631,8 @@ describe('imperial', () => {
           edges: [['a', 'c']],
         });
 
-        const game = new Imperial(board);
+        const logger = new Logger('silent');
+        const game = new Imperial(board, logger);
```
GitGubM8, line 4635: Hm should we reinstate the `unittest` logger scope then if it turns out that failing tests become quite noisy

GitGubM8: Also we'll probably have to update all the `new Imperial`s sadly, probably worth also reinstating the `newImperial` helper again

Thrillberg: I prefer something like `silent` to clarify that it is not just meant for tests but for particular tests (ie, those that purposefully send invalid actions though). In other cases, the tests were just wrong and I've fixed them. Having those cases fall through to `console.error()` was actually very helpful in resolving the error! But as of this commit, the tests are no longer noisy.

Thrillberg: >Also we'll probably have to update all the new Imperials sadly, probably worth also reinstating the newImperial helper again

Why? things are passing fine here, no?

GitGubM8: When the other unittests do fail, it'd be noisy again wouldn't it?

As in eventually, if the unittests are doing their job, they're going to fail at some point in time, which might result in the same issue

Thrillberg: Ideally they won't. Or if they do we'll want to actually fix the tests rather than silence the failures, right?

GitGubM8: Fair, just worried about the visibility of the error if and when that happens

For example in this PR's case, technically we didn't fix the issue because the invalid logs are still here, yet we're suppressing the error logs because it isnt really worth the effort, the logs don't contribute much and it's an acceptable "fail"

As long as you're aware of the potential, that's fine. Just food for thought. Don't particularly care right now since everything's working smoothly

GitGubM8 (approved):
Lgtm!

app/javascript/lib/imperial.test.js
Thrillberg: Screenshot 2023-03-28 at 3 32 03 PM
This deals specifically with invalid moves. They are necessary to run these tests because it is testing that the system will ignore them. In all other cases, though, we want to see the logs.

GitGubM8: Ahh I see, I was mistaken about the test itself then, gotcha, carry on!

PR #734, opened 2023-03-29
Fix challenger and incumbent in pre-existing fights

Thrillberg:
@GitGubM8 Hopefully this doesn't interfere with your current PR too much!


GitGubM8 (approved):
Beautiful, doesn't look like it conflicts with anything I have actually

PR #735, opened 2023-03-29
Renamed imperial.js -> ImperialGameCoordinator

GitGubM8:
## Changes:
- Also moved some tests file into new folder path `javascript/Tests/IntegrationTests`


Thrillberg:
I'm not sure if I'd call the `imperial.test.js` tests integration tests. They aren't really unit tests either, but they don't, for example, deal with the front end, which is usually what I'd think of an integration test as doing. I'd be more comfortable calling them unit tests or maybe something else in between?

Also, I think they should still live inside of `/lib` since the code in `/src` is unrelated to those tests. Maybe rename `/lib` to something more descriptive like `/gameLogic`? and `/src` to something like `/web`? Not sure, but the current names are not too helpful!

GitGubM8:
Ah I see what you're saying, I agree that if it deals with across different concerns then they definitely should be called intergration-tests.

Pretty minor, but I guess the difference in my definition is that since the tests here often deal with interact with usecases in one go, and therefore qualifies as integration-tests,

whereas a test that deals with only a specific usecase would count as a unit-test?

For example, I guess if I were a bit stringent, I'd probably be writing separate unittest right now for specifically the `AvailableSlots`, `SlotDistanceCost`, and `MoveToSlot` logic (altho for `MoveToSlot` you could argue that it deals with multiple usecases as well, which I'd have no good retort)

Again, happy to name it whatever you think is good

GitGubM8:
Also should I rename the folders in another PR? Worried that doing so would also cause thousand of line changes

Thrillberg:
>the difference in my definition is that since the tests here often deal with interact with usecases in one go, and therefore qualifies as integration-tests

Yeah I hear you. I think before they were unit tests because the use cases you're splitting off didn't exist. So now they're becoming less clearly unit-oriented and more like "system" tests? I don't love `system` as a word for tests because it's so vague but it might be best here. Maybe we should put some explanation on what we mean by "system" in the README. No need to belabor this toooooo much but just wanted to throw my thoughts out there. Do what you think is best!

>I'd probably be writing separate unittest right now for specifically the AvailableSlots, SlotDistanceCost, and MoveToSlot logic

Makes sense. I think it's fine to skip it for now and keep the testing at this kind of in-between unit and integration level for now.

>Also should I rename the folders in another PR? Worried that doing so would also cause thousand of line changes

I'm only seeing +20/-22 lines of change so I think this is fine!
Screenshot 2023-03-29 at 2 59 26 PM

GitGubM8:
I did warn you 😢 
image

Thrillberg (approved):
Oh I see yes with all the linting this is huge. The meaningful changes seem minimal so I think it's fine

PR #736, opened 2023-03-29
Fix color of details on EU, CN & IN flags, and fix BR flag clip

Thrillberg:
This resolves that issue we looked at together awhile ago. I think flags are looking better! The screenshots are from two different games because after fixing it I couldn't repro the problem locally _even after_ switching back to `main`. 🧐

**Before**
Screenshot 2023-03-29 at 2 53 34 PM

**After**
Screenshot 2023-03-29 at 2 51 36 PM


app/javascript/src/components/flags/INFlag.vue
```diff
@@ -6,133 +6,160 @@
     :height="height"
   >
     <filter id="grayscale">
-      <feColorMatrix type="matrix" values="0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0 0 0 1 0"/>
+      <feColorMatrix
```
Thrillberg, line 9: This file is mostly just linting :/

GitGubM8 (approved):
Wow, night and day

PR #744, opened 2023-04-01
ImperialGameCoordinator no longer casted into proxy

GitGubM8:
## Comments:
- Having the game being casted into proxies was still causing issues eventually down the line w.r.t. private fields, forcing it to always stay as obj


Thrillberg (approved):
Cool! Looks good to me. We probably want to address the GH Actions warnings at some point but it's not blocking. Thanks!

PR #745, opened 2023-04-01
Undo

GitGubM8:
## Depends on:
- https://github.com/Thrillberg/imperial/pull/744

## Comments:
- If I understand properly, the existing implementation of `undo` basically deep copies every value, and replaces all of it when `undo`ing
- It's very nice for modularity, requiring no special logic even if other components change, altho it seems like it might be heavy in terms of memory needed as the game drags on, since it tracks all the states since the beginning

- Unfortunately, with a more OO approach, at least currently, not everything can be cloned as easily. The approach I've opted for is to have an undo-stack that stores lambdas that are able to undo specifically that move, but would require every usecase that introduces changes to manually log the changes they've made.


app/javascript/Domain/ImperialGameCoordinator.js
```diff
@@ -117,6 +118,10 @@ export default class ImperialGameCoordinator {
       return;
     }
 
+    if (action.type !== 'undo') {
+      this.#game.addUndoCheckpoint();
```
Thrillberg: Just so I understand this, we add an empty array for every action and _some_ of those arrays will be populated with undo operations if the player did an undo on an action? But in most games most of those arrays will remain empty forever?

Thrillberg (commented):
>require every usecase that introduces changes to manually log the changes they've made

I'm debating about whether we should make this change. Your understanding of the current operation of `undo` is spot on. I've always thought there is benefit to being able to log an "undo" and then the game just recalculates _without_ the prior, undone move. This way we don't have to think about the ramifications of every move AND every move in reverse. Is there benefit to having the codebase know how to reverse every action? I'd be curious about your thoughts on that.

app/javascript/Domain/ImperialGameCoordinator.js
GitGubM8: In most of the time, the undo stack will be populated with operations, and I don't think there's ever a case where any stack-frame will be empty, I haven't gotten around to implementing them yet since we'd want to keep the PRs small

GitGubM8:
Having tried refactoring both the turn order and the taxation system out, I was unable to progress much further without eventually coming back to make sure the undo still worked for players, and needed to constantly translate changes from the old entities-models to the new entities-model, which was a major pain and effectively bloated the existing undo action anyway

I like the current system better, but as we expand the refactor more, Im also not too keen on having to store in memory a duplication of every object on every move

While it does incur some overhead when implementing individual actions, since now we also have to write the undo operation, I think overall it's still a slight benefit (at least for now), since

A. there'll be less memory usage, the diff between each move is always less than the whole game (and that may or may not speed up game processing time when loading-in or rolling back to view older turns)
B. for the duration of the refactor, I won't have to worry about constantly translating entity-models back and forth from the old system to the new system
C. I don't have to immediately implement new methods for every entity to support cloning (although we'll still have to eventually get to it when implementing AIs)

GitGubM8:
There's also another arugment we could make about how the undo system should be entirely separate from the game system itself tbh, which I'd actually like more

For example, when we do get around to writing AIs, I wouldn't want them to ever undo their moves; Having it stored in the game only weighs the system down every time it expands the search space

Edit:
Now that I think about it, let me pull this system out, I think there's a slightly better design for this, altho it'll still be an undo-stack

GitGubM8:
[51ee24c](https://github.com/Thrillberg/imperial/pull/745/commits/51ee24cc086c597053f53f41f0de1ff558053a92)

I've pulled the undo system out so it'll be easier making future changes towards the undo system in general, be it storing a copy of every game state vs only storing the diffs

app/javascript/Domain/ImperialGameCoordinator.js
```diff
@@ -117,6 +121,10 @@ export default class ImperialGameCoordinator {
       return;
     }
 
+    if (action.type !== 'undo') {
+      this.#undoHistory.addUndoCheckpoint();
```
Thrillberg: This will add a possible undo to every action, right? The way undo currently works is that it will undo an entire turn at a time and not just one action at a time. Are we changing behavior with this?

Thrillberg (approved):
Ok I think I'm coming around to this way of handling undos. There's definitely something more organized about defining each action's reversal, rather than the somewhat hacky-feeling reversion to some old state. Let's go for it!

I have one question in here which isn't a blocker so I'll approve for now but just wanted to make sure it was on your radar.

app/javascript/Domain/ImperialGameCoordinator.js
GitGubM8: Ah, good catch, let me fix that (not sure why it didn't fail for the games on my local machine)

GitGubM8: [1c3054e](https://github.com/Thrillberg/imperial/pull/745/commits/1c3054eee2299941ea44ad0d96cb6f5389718f08)

Moved into `setOldState` which is also renamed as `#addUndoCheckpoint` now

GitGubM8:
Also tested with bold-frog imported game, looks good

PR #756, opened 2023-04-03
Send latest game state to Redis with each tick

Thrillberg:


GitGubM8:
Pretty sure you know already, but

Keep in mind that there's going to be data structural changes through refactors down the line, which can conflict with the game state previously cached within redis

Migration seems like a pain for something that's only meant for beautifying the site; Flushing the cache everytime we deploy could be something worth looking into

(alternatively, as mentioned IRL chat, while it doesnt seem like the most efficient way, caching a screenshot is a foolproof catch-all solution too)

Thrillberg:
>there's going to be data structural changes through refactors

Yeah we'll have to keep on top of those for this feature. It might be worth thinking more purposefully about the "game state" as a canonical entity. One trouble I'm running into is converting our JS `Map`s (and presumably `Set`s) into JSON for transmitting to the cache. So if we have a canonical game state set of keys, we probably want to exclude any Maps and Sets.

>Flushing the cache everytime we deploy

This is how I'm anticipating it will work. Redis flushes every time the server is restarted (which happens on deploy). The challenge then will be to repopulate the cache on initialization.

>caching a screenshot is a foolproof catch-all solution

That seems less interesting to me (could be a totally biased perspective) than handling these caching issues this way. Not sure which way is "better" but given that this is a side project with no financial impact anywhere, I'm inclined to do what I prefer 🤣 

Thanks for raising these issues! Really helpful!

Thrillberg:
I think we should put a hold on this because the Map and Set problem is going to need a bit of a bigger refactor. Not sure it's worth it at this point. But this is a useful SO answer, for future reference: https://stackoverflow.com/questions/29085197/how-do-you-json-stringify-an-es6-map

PR #759, opened 2023-04-04
Bonds

GitGubM8:
## Comments:
- Doesn't actually refactor anything yet, just new bond entities and bond purchasing use cases
- Should be ready to refactor `Auction.js` next

### DIFF
--- a/app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
+++ b/app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
@@ -1,6 +1,6 @@
 const InsufficientTreasuryError = class extends Error {
   constructor(nation, value) {
-    super(`${nation} does cannot afford ${Math.abs(value)}`);
+    super(`${nation} cannot afford ${Math.abs(value)}`);
 
     this.name = 'InsufficientTreasuryError';
   }
--- a/app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
+++ b/app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
@@ -4,7 +4,7 @@ const InsufficientTreasuryError = class extends Error {
 
     this.name = 'InsufficientTreasuryError';
   }
-}
+};
 
 export default class AdjustTreasury {
   static get InsufficientTreasuryError() {
--- a/app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
+++ b/app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
@@ -1,0 +1,28 @@
+const InsufficientTreasuryError = class extends Error {
+  constructor(nation, value) {
+    super(`${nation} does cannot afford ${Math.abs(value)}`);
+
+    this.name = 'InsufficientTreasuryError';
+  }
+}
+
+export default class AdjustTreasury {
+  static get InsufficientTreasuryError() {
+    return InsufficientTreasuryError;
+  }
+
+  static changeBy(nation, amount, undoHistory) {
+    if (nation.treasury + amount < 0) {
+      throw new InsufficientTreasuryError(nation, amount);
+    }
+
+    if (undoHistory) {
+      const originalAmount = nation.treasury;
+      undoHistory.pushUndoOperation(() => { nation.treasury = originalAmount; });
+    }
+
+    if (amount) {
+      nation.treasury = Math.max(0, nation.treasury + amount);
+    }
+  }
+}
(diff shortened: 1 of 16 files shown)


app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
```diff
@@ -0,0 +1,28 @@
+const InsufficientTreasuryError = class extends Error {
+  constructor(nation, value) {
+    super(`${nation} does cannot afford ${Math.abs(value)}`);
```
Thrillberg: Small grammar error here. Remove `"does"`.

app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
```diff
@@ -0,0 +1,28 @@
...
+  static get InsufficientTreasuryError() {
+    return InsufficientTreasuryError;
+  }
+
+  static changeBy(nation, amount, undoHistory) {
+    if (nation.treasury + amount < 0) {
+      throw new InsufficientTreasuryError(nation, amount);
+    }
+
+    if (undoHistory) {
```
Thrillberg, line 19: Once we have the new undos fully in place will we remove conditionals like this one?

app/javascript/Domain/UseCases/Nations/Bonds/AffordableBonds.js
```diff
@@ -0,0 +1,37 @@
...
+    const sameNationBonds = [];
+    for (const bond of investor.bonds) {
+      if (bond.nation === nation) {
+        sameNationBonds.push(bond);
+      }
+    }
+
+    for (const bond of nation.unsoldBondsByInterestValue.values()) {
+      if (investor.cash >= bond.cost) {
+        yield bond;
```
Thrillberg, line 20: Very small, incidental question but is there a reason we duplicate the code from `bondsPurchasableInFull` here rather than call it and include that return value with the additional bonds that can be upgraded for?

app/javascript/Domain/UseCases/Players/AdjustCash.js
```diff
@@ -1,12 +1,28 @@
+const InsufficientCashError = class extends Error {
+  constructor(investor, value) {
+    super(`${investor} does cannot afford ${Math.abs(value)}`);
```
Thrillberg: Same grammar error as above.

Thrillberg (approved):
Looks good! Nothing really earth-shattering in my comments here but I'm having a fun time reading through this code. I've never used an architecture quite like this so it's interesting to see how it's reflected in Imperial.

app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
GitGubM8: Oh, those are necessary and permanent

In JS it's not that easy to indicate intent without overloading or typing, but `undoHistory` here is optional
With that, users of the use case aren't forced to always have their operation be undo-able

An example of that is when actually undoing the operation, the logic is highly similar, but we don't write to the undo-stack (IE. you can't undo an undo/ctrl-y)

app/javascript/Domain/UseCases/Nations/Bonds/AffordableBonds.js
GitGubM8: Mostly to avoid having to overthink the logic on how to make sure we're not yielding duplicate bonds

In other to avoid duplicates, even if we yield those in `bondsPurchasableInFull` directly, we'd still have the wrap the logic below (the upgrade logic) in an `if (cannot purchase in full)`, which is about the same

```
for (const bond of bondsPurchasableInFull(nation, investor)) {
  yield bond;
}

for (const bond of nation.unsoldBondsByInterestValue.values()) {
  if (investor.cash < bond.cost) {
    // yield upgradable bonds logic
  }
}
```

vs
 
```
for (const bond of nation.unsoldBondsByInterestValue.values()) {
  if (investor.cash < bond.cost) {
    // yield upgradable bonds logic
  } else {
    // yield bond directly
  }
}
```

up to you idrc either way

app/javascript/Domain/UseCases/Players/AdjustCash.js
GitGubM8: [568484e](https://github.com/Thrillberg/imperial/pull/759/commits/568484ee4b471a4c01941abf4ef564ea7f2c4f18)

app/javascript/Domain/UseCases/Nations/AdjustTreasury.js
GitGubM8: [568484e](https://github.com/Thrillberg/imperial/pull/759/commits/568484ee4b471a4c01941abf4ef564ea7f2c4f18)

PR #764, opened 2023-04-04
Nation Turn

GitGubM8:
# Depends On:
- https://github.com/Thrillberg/imperial/pull/759

## Changes:
- New GiveNationTurn usecase

## Comments:
- I was actually able to clean up the entire `Auction.js`, but the references from `ImperialGameCoordinator` was a bit too much
- Going to just commit another dangling use case for now

PR #766, opened 2023-04-06
Player & Swiss bank

GitGubM8:
## Depends on:
- https://github.com/Thrillberg/imperial/pull/764

## Changes:
- Minor refactor of new Player class
- Added SwissBanker to AbstractImperialGame
- Simplified AbstractImperialGame


GitGubM8:
Hm some cypress errors about access error

Looking through the stack, I think it's complaining about objects being casted into a proxy instead, and therefore no longer counts as being the same class technically, thus violating access modifiers

Not familiar with how to fix cypress, but for what it's worth, I managed to fix similar issues in `Game.vue` using `markRaw`

GitGubM8:
There we go, finally fixed the cypress issues, glad it was complaining since it was a legitimate issue

That being said still don't fully comprehend how the frontend calls other frontend modules

Thrillberg (approved):
This looks good to me! I didn't do a super thorough review but I liked what I saw. Thanks for your patience with rebasing and all that while I was reworking the FE! 🎉

PR #792, opened 2023-04-21
Auction Coordinator

GitGubM8:
## Comments:
- No real changes, ImperialGameCoordinator is proving to be pretty stubborn
- New AuctionCoordinator, more UseCases


app/javascript/Domain/AuctionCoordinator.js
```diff
@@ -0,0 +1,67 @@
...
+  #gameCoordinator;
+
+  #nationToAuction;
+  #participatingPlayersInOrder;
+
+  #auctionPlayerIndex = -1;
+
+  #isComplete = false;
+
+  constructor(gameCoordinator, nationToAuction, participatingPlayersInOrder) {
```
GitGubM8, line 15: Instead of performing all the auctions at once, each AuctionCoordinator instance will represent a single round of auction for 1 country only.

This flexibility should allow `IGC` to reuse the same code for even individual bond purchases + swiss bankers

GitGubM8:
I think the majority of the commits stem from the branch branching off previous unmerged branches. For the next one I'll branch off `main` to avoid the long history

GitGubM8:
Going to merge this since this should be pretty harmless, and I want to branch off main with less atrocious of a commit history (squash & merge)

PR #804, opened 2023-05-09
modified black sea and caspian sea fill color

### LINKED ISSUE #700: Recolor Black Sea and Caspian Sea
In the beginning, I was slightly confused about the white blots, which actually represented the Black Sea and the Caspian Sea

image

kchender:
Black and Caspian Seas were white in Imperial 2030, now they are blue.


Thrillberg (approved):
Looks great to me! Thanks! Hopefully these don't get confusing with the color of the EU provinces but if they do we can tweak it.

